PR #23, opened 2015-05-29
Load libc using NativeLibrary.getProcess()

cladden:
libc has already been mapped into the address space for you by the
runtime loader, ld.so, long before the jvm's main() entry point was
invoked.

The correct way in Java to get dlsym access to the existing runtime
linker symbol table is:

Native.register(NativeLibrary.getProcess())


brettwooldridge:
Tnanks!

PR #29, opened 2015-08-11
Add destroyForcibly() method

rdicroce:
On *nix OSes, the NuProcess#destroy() method does not (cannot) guarantee that the process will terminate. It sends SIGTERM, which could be ignored by the target process. To deal with this, I have added a destroyForcibly() method similar to the one added to the JDK's Process class in Java 8. Calling destroyForcibly() sends SIGKILL, which is guaranteed to terminate the process. On Windows, destroy() and destroyForcibly() do the same thing, since TerminateProcess is already guaranteed to actually terminate the process.

### DIFF
--- a/src/example/java/com/zaxxer/nuprocess/example/SshExample.java
+++ b/src/example/java/com/zaxxer/nuprocess/example/SshExample.java
@@ -71,7 +71,7 @@ public class SshExample
 
         public void close()
         {
-            nuProcess.destroy();
+            nuProcess.destroy(false);
         }
 
         //Send key event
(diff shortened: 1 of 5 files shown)


brettwooldridge:
It's a reasonable change.  However, I'd prefer not to increase the API surface area (I'd rather break the ABI).  So, rather than introducing ``destroyForcibly()``, alter ``destry()`` to take a parameter ... ``destroy(boolean force)``.

rdicroce:
I'll adjust the PR when I have time, probably next week. The only reason I didn't add a boolean parameter was to preserve API compatibility. There's going to be some confusion after you release a new version with this change merged, because people are going to have compile errors and aren't going to be sure what the previous behavior was.

brettwooldridge:
We'll document the change, and following semantic versioning will bump the major API version number to express ABI incompatibility.

rdicroce:
I've adjusted the PR per your comments. Let me know what you think.

brettwooldridge:
Looks good, thanks!

PR #30, opened 2015-08-12
New classes NuCharsetDecoder and NuCharsetEncoder to handle conversion of bytes to string data

bhamiltoncx:
Thanks for the nice library! I'm an engineer on the Buck Build team at Facebook. Buck currently has some pretty crummy and non-performant abstractions around process management, and `NuProcess` is just what the doctor ordered.

When using `NuProcess`, I found it's very common to want to use `CharsetDecoder` to decode stdout/stderr bytes provided to `NuProcessHandler` into `String` data (and to encode String data to bytes for stdin).

Right now, it's pretty inconvenient to do that: if you don't consume all the bytes in the `ByteBuffer` passed to `onStdout()` (say, because a UTF-8 code point boundary straddles two reads), the buffer is wiped before the next invocation. That means you have to do your own byte buffering in order to decode the data.

Since this seems like a pretty common use-case, this pull request adds a new concrete implementation of `NuProcessHandler` called `NuProcessDecoder` which handles streaming decoding and encoding of stdin, stdout, and stderr bytes to and from Java strings.

Previously, stdout and stderr shared a single buffer in `BasePosixProcess`. After this diff, they are separate buffers.

In addition, we now optionally support keeping unused data in the buffers after `onStdout()` or `onStderr()` returns, using `ByteBuffer.compact()` to move the unused data to the front of the buffer so it can be consumed on the next invocation of that method. This new behavior is off by default, and can be enabled by returning `true` from `NuProcessHandler.shouldCompactBuffersAfterUse()`.

Since we can now have unused data in the stdout and stderr buffers at exit, I added two new callbacks `onPreExitStdout()` and `onPreExitStderr()` to allow callers one last chance to consume stdout/stderr data at process exit time. This is necessary for `CharsetDecoder` to finalize any decoding to Unicode.

I included docs and unit tests, but I only tested this on OS X 10.10. I think it ought to work on Windows as well, but I hope you (or CI) can test that for me.


bgertzfield:
Another possibility here is to have a general-purpose single stream decoder / encoder which can be optionally hooked in to separately decode or encode stdout, stderr, or stdin.

brettwooldridge:
Thanks for the pull request.  Give me a day to look it over.  Off the top of my head...

It seems that *shouldCompactBuffersAfterUse* can be implicit based on whether *after* the call to ``onStdout()/onStderr()`` return the buffer has remaining (run read) content.

brettwooldridge:
I'd rather not increase the "surface area" of the API.  I'd rather break the ABI and have a cleaner interface in the end.  I have two alternative suggestions to that end.

1. Call ``onStdout()`` and ``onStderr()`` in the exit path if the ByteBuffers have content remaining.  Possibly changing the signature to ``onStdout(ByreBuffer buf, boolean closed)`` if knowing it is the final call is important.

or

2. Change the ``onExit()`` signature to ``onExit(int exitCode, ByteBuffer outBuf, ByteBuffer errBuf)``.

bgertzfield:
> It seems that shouldCompactBuffersAfterUse can be implicit based on whether after the call to onStdout()/onStderr() returns the buffer has remaining (unread) content.

I was originally going to do that, but I was worried about breaking backwards compatibility. If that's not a problem, it's easy to change.

> I'd rather not increase the "surface area" of the API.

Good suggestions. Will give it a go. I think I'll also decouple decoding of stdout from stderr and make the encoders/decoders pluggable into any of stdin, stdout, or stderr.

bhamiltoncx:
Thanks for all the feedback! I've integrated the following changes:

- Remove `onPreExitStdout` and `onPreExitStderr`.
- Replace with boolean `closed` on `onStdout` and `onStderr`.
- No more passing `null` to `onStdout` or `onStderr` to indicate closed condition.
- Split stdin, stdout, and stderr encoding/decoding into separate `NuProcessDecoder` and `NuProcessEncoder`.
- New abstract helper class `NuAbstractProcessEncodingHandler` to combine stdin, stdout, and stderr encoding/decoding.
- Make compacting buffer implicit based on whether buffer has anything remaining.
- Correctly pass remaining stdout or stderr bytes if they are closed before process exit.

bhamiltoncx:
By the way, I noticed javadoc is no longer building cleanly in master:

https://gist.github.com/bhamiltoncx/b1d783b3ed44df2baa67

brettwooldridge:
@bhamiltoncx I'm basically happy with the result, thanks.  I do have a few tweaks to request (how could it be otherwise for such a big pull?).

Can you move the encoding/decoding classes into a new ``com.zaxxer.nuprocess.codec`` package?  Should be just a few clicks in your IDE.

Also, I probably set a bad precedent with the ``NuAbstractProcessHandler`` class naming.  Can you rename the classes as:
```
NuAbstractProcessEncodingHandler → NuAbstractCharsetEncodingHandler
NuAbstractProcessDecodingHandler → NuAbstractCharsetDecodingHandler
NuProcessEncodingHandler → NuCharsetEncodingHandler
NuProcessDecodingHandler → NuCharsetDecodingHandler
NuProcessEncoder → NuCharsetEncoder
NuProcessDecoder → NuCharsetDecoder
```
Lastly, I have a small number of comments to add inline...

Thanks again for the contribution,

src/main/java/com/zaxxer/nuprocess/NuAbstractProcessEncodingHandler.java
```diff
@@ -0,0 +1,179 @@
...
+   * encoded and decoded string data.
+   *
+   * @param stdinEncoder The {@link CharsetEncoder} with which to encode stdin bytes
+   * @param stdoutDecoder The {@link CharsetDecoder} with which to decode stdout bytes
+   * @param stderrDecoder The {@link CharsetDecoder} with which to decode stderr bytes
+   */
+  protected NuAbstractProcessEncodingHandler(
+      CharsetEncoder stdinEncoder,
+      CharsetDecoder stdoutDecoder,
+      CharsetDecoder stderrDecoder) {
```
brettwooldridge: These three args are short enough to include on one line (96).

src/main/java/com/zaxxer/nuprocess/NuProcessEncoder.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Creates an encoder which uses a {@link CharsetEncoder} to encode input data.
+   *
+   * @param handler {@link NuProcessEncoderHandler} called back with a string buffer
+   *                into which the caller writes string data to be written to stdin
+   * @param encoder {@link CharsetEncoder} used to encode stdin string data to bytes
+   */
+  public NuProcessEncoder(
+      NuProcessEncoderHandler handler,
+      CharsetEncoder encoder) {
```
brettwooldridge: These two args are short enough to include on one line (61).

src/main/java/com/zaxxer/nuprocess/NuProcessHandler.java
```diff
@@ -17,6 +17,7 @@
 package com.zaxxer.nuprocess;
 
 import java.nio.ByteBuffer;
+import java.nio.charset.CharsetDecoder;
```
brettwooldridge: Unused import.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -383,23 +397,32 @@ public void readStdout(int availability)
       try {
          if (availability < 0) {
             outClosed = true;
-            processHandler.onStdout(null);
+            outBuffer.flip();
+            processHandler.onStdout(outBuffer, true);
             return;
          }
          else if (availability == 0) {
```
brettwooldridge: I now worry about this conditional.  If the developer does not consume all of the data, as they were required to in the past, they may get into trouble of the buffer fills completely.  They will never be called again to consume data.

src/main/java/com/zaxxer/nuprocess/windows/ProcessCompletions.java
```diff
@@ -237,7 +237,7 @@ private void queueWrite(final WindowsProcess process)
 
    private void queueRead(final WindowsProcess process, final WindowsProcess.PipeBundle pipe, final int stdX)
    {
-      if (NuKernel32.ReadFile(pipe.pipeHandle, pipe.bufferPointer, NuProcess.BUFFER_CAPACITY, null, pipe.overlapped) == 0) {
+      if (NuKernel32.ReadFile(pipe.pipeHandle, pipe.bufferPointer.share(pipe.buffer.position()), pipe.buffer.remaining(), null, pipe.overlapped) == 0) {
```
brettwooldridge: What happens if pipe.buffer.remaining() is 0?

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Good catch. We'll now throw a `RuntimeException()` if the buffer has 0 capacity remaining after compacting.

src/main/java/com/zaxxer/nuprocess/windows/ProcessCompletions.java
bhamiltoncx: I think it'll be handled by the above check I added after `onStdout()` / `onStderr()` return, but I added another check here and we now throw `RuntimeException` in this case.

src/main/java/com/zaxxer/nuprocess/NuProcessHandler.java
bhamiltoncx: Removed.

src/main/java/com/zaxxer/nuprocess/NuProcessEncoder.java
bhamiltoncx: Fixed.

src/main/java/com/zaxxer/nuprocess/NuAbstractProcessEncodingHandler.java
bhamiltoncx: Fixed.

bhamiltoncx:
> Can you move the encoding/decoding classes into a new com.zaxxer.nuprocess.codec package? Should be just a few clicks in your IDE.

Done. (Actually, I don't use an IDE, but I booted up IntelliJ just for that. ;)

> Also, I probably set a bad precedent with the NuAbstractProcessHandler class naming. Can you rename the classes as:

Done.

bhamiltoncx:
```
Running com.zaxxer.nuprocess.InterruptTest
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

CI build failed: is this test known to be flaky? Looking at the source, it seems possible.

brettwooldridge:
Thanks for the nice contribution

bhamiltoncx:
Glad to help!

itaifrenkel:
@brettwooldridge Would it be possible to increment the pom.xml version and release to maven repo ? thanks :)

brettwooldridge:
Done.  Release is published to sonatype repository.  It usually takes 3-4 hours to be indexed by the search engine, but you might be able to use the dependency right now.

PR #31, opened 2015-09-04
Performance improvements for OS X kqueue / kevent

bhamiltoncx:
I noticed the way `ProcessKqueue` was using `kevent()` was pretty inefficient: it was repeatedly adding and removing the stdio monitoring events to the `kqueue`, and it was handling only one event at a time.

Since each time we call `kevent()` it requires a round-trip to the kernel, it's important to call it as few times as possible. The most efficient way to use `kevent()` is to:

1. Re-use events by registering them without the `EV_ONESHOT` flag whenever possible
2. Pass in an array of kevents to read as many events as possible at the same time
3. Rely on the OS automatically invalidating stdio-monitoring events when the fd is closed

This diff replaces the single `Kevent triggeredEvent` with a new `KeventArray` holding 64 events which is re-used repeatedly to avoid reallocating the array. Now, each `Kevent` just contains a `Pointer` into a `KeventArray`.

(We probably should rewrite all this code to use a JNA `Structure` but I didn't feel like doing such a big change here.)

For the closeQueue / wantsWrite queues, I also made a tweak to use the efficient `BlockingQueue.drainTo()` API which removes multiple pending items in one call instead of repeatedly polling the queue.

Finally, I removed the `BlockingQueue<Kevent> keventPool`, since it's no longer needed.


src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
```diff
@@ -33,13 +36,12 @@
...
 {
-   private static final int KEVENT_POOL_SIZE = 16;
+   private static final int NUM_KEVENTS = 64;
    private static final TimeSpec timeSpec;
    private static final int JAVA_PID;
 
    private volatile int kqueue;
-   private Kevent triggeredEvent;
-   private BlockingQueue<Kevent> keventPool;
+   private ThreadLocal<KeventArray> keventArrays;
```
brettwooldridge: I have had no end of troubles with ThreadLocals and application containers, like Tomcat, that sniff for ThreadLocal leaks...

https://github.com/brettwooldridge/HikariCP/issues?utf8=✓&q=is%3Aissue+is%3Aclosed+threadlocal+

[Classloader Leaks](http://www.slideshare.net/mjiderhamn/classloader-leaks/28)

src/main/java/com/zaxxer/nuprocess/osx/KeventArray.java
```diff
@@ -0,0 +1,62 @@
...
+   }
+
+   public Kevent get(int i) {
+      return kevents.get(i);
+   }
+
+  public int size() {
+    return size;
+  }
+}
```
brettwooldridge: Here is what I'm worried about.  I cheated.  I'm not sure that each Kevent is actually 32 bytes.  I think I studied it and decided that it was *less than 32 bytes*.  It could be as small as 16 bytes...

And because I only read one event at a time, there was no issue.  But when passing an array of kevent objects to the native side the size needs to be exact, because only the address of the first kevent is passed.

Back in the day, Kevent *was* [a structure](https://github.com/brettwooldridge/NuProcess/blob/a10d761f16688936338e091df9d42af0684a9b23/src/main/java/org/nuprocess/osx/Kevent.java).  But I think my mapping wasn't quite right, and I would occassionally get memory access violations.

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
```diff
@@ -192,59 +261,38 @@ else if (WIFSIGNALED(status)) {
...
-      if (closeQueue.isEmpty()) {
-         return;
-      }
-
-      OsxProcess process;
-      while ((process = closeQueue.poll()) != null) {
-         process.stdinClose();
-      }
+     List<OsxProcess> processes = new ArrayList<OsxProcess>();
+     closeQueue.drainTo(processes);
```
brettwooldridge: The trouble with ``drainTo()`` is that it is not atomic.  The JavaDoc states: ``the behavior of this operation is undefined if the specified collection is modified while the operation is in progress.``  In practical terms, it is possible for a process to be in both the drained list and yet still remain in the queue if an element is added during the ``drainTo()`` operation.

The efficiency gain here is likely not measurable in real world use, so I recommend leaving this method as is.

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
```diff
@@ -192,59 +261,38 @@ else if (WIFSIGNALED(status)) {
...
-         return;
-      }
-
-      OsxProcess process;
-      while ((process = wantsWrite.poll()) != null) {
-         int fd = process.getStdin().get();
-         if (fd != -1) {
-            queueEvent(process.getStdin().get(), Kevent.EVFILT_WRITE, Kevent.EV_ADD | Kevent.EV_ONESHOT, 0, new Pointer(process.getPid()));
+      List<OsxProcess> processes = new ArrayList<OsxProcess>();
+      wantsWrite.drainTo(processes, NUM_KEVENTS);
```
brettwooldridge: Same as above.

src/main/java/com/zaxxer/nuprocess/osx/KeventArray.java
brettwooldridge: In summary, I think if you intend to call the ``kevent(kqueue, null, 0, keventArray.getPointer(), NUM_KEVENTS, timeSpec)`` method requesting multiple events at once, the size of the Kevent allocation must be exact.  It is not necessary to return to a structure, but at a minimum validate the size.  

Maybe 32-bytes is correct (I can't remember)...

src/main/java/com/zaxxer/nuprocess/osx/KeventArray.java
```diff
@@ -0,0 +1,62 @@
...
+      this.size = size;
+      long memory = Native.malloc(32 * size);
+      pointer = new Pointer(memory);
+      kevents = new ArrayList<Kevent>(size);
+      for (int i = 0; i < size; i++) {
+         kevents.add(new Kevent(pointer.share(i * 32)));
+      }
+   }
+
+   public void free()
```
brettwooldridge: free() is never called.

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
```diff
@@ -84,15 +94,52 @@ public void registerProcess(OsxProcess process)
 
       pidToProcessMap.put(pid, process);
 
-      // Listen for process exit
-      queueEvent(process.getPid(), Kevent.EVFILT_PROC, Kevent.EV_ADD | Kevent.EV_ONESHOT, Kevent.NOTE_EXIT | Kevent.NOTE_EXITSTATUS | Kevent.NOTE_REAP,
-                 pidPointer);
-      // Listen for self-generated signal to tell processor to wake-up and handling pending stdin requests
-      queueEvent(LibC.SIGUSR2, Kevent.EVFILT_SIGNAL, Kevent.EV_ADD, 0, pidPointer);
+      KeventArray keventArray = keventArrays.get();
```
brettwooldridge: This is going to get called by the user's thread -- the one calling ``NuProcessBuilder.start()`` -- which is going to create a new ThreadLocal array of 64 kevents.  This is really inefficient compared to the kevent pool that was used.  Especially if the threads launching processes are one-shot.

We know we only need 4 kevent objects here, so why not construct what we need and skip the ThreadLocal altogether?

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
brettwooldridge: This not an issue if the only interaction is this ThreadLocal is by a NuProcess managed thread (see comments on ``registerProcess(OsxProcess process)`` below).

However, the [``BaseEventProcess.run()``](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java#L61) method *must* then have a ``finally`` that calls a shutdown/cleanup method so that:

 * ``KeventArray.free()`` gets called, and
 * ``keventArrays.remove()`` is called to remove the ThreadLocal

bhamiltoncx: Totally agreed, the shut down path was broken before as well. I'll clean it
up.

I'm happy to convert to use a JNA Structure instead of the existing logic
for the kevent size.

I don't feel too strongly about the drainTo() call, but I did check the
source and it is indeed atomic.
2015年9月4日(金) 23:21 Brett Wooldridge <notifications@github.com>:

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
brettwooldridge: @bhamiltoncx You're right, ``ArrayBlockingQueue`` implements an atomic ``drainTo()``.  I'm fine leaving this, but put a comment to the effect of:
```java
// drainTo() is known atomic for ArrayBlockingQueue
```

src/main/java/com/zaxxer/nuprocess/osx/KeventArray.java
bhamiltoncx: Ah, I see what was wrong with the old code.

struct kevent is declared under `#pragma pack(4)`, so JNA needs to be manually told what the alignment of each member of the struct is. I'll send a fix.

bhamiltoncx: Turns out this isn't in fact needed. I confirmed the default alignment is correct with a small C program using `offsetof()` to compare to `Structure.getOffset()`.

I noticed your original code was directly passing `Kevent[]`, but that doesn't seem to work with JNA 4.1.0 and direct mapping of native methods. So I changed it to pass `Pointer` and manually call `Structure.read()` and `Structure.write()`.

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/osx/ProcessKqueue.java
bhamiltoncx: Fixed.

src/main/java/com/zaxxer/nuprocess/osx/KeventArray.java
bhamiltoncx: Removed method entirely.

bhamiltoncx:
OK, I think I fixed all the issues. Thanks for the thorough review!

PR #32, opened 2015-09-05
Process cwd support for Mac, Linux, and Win32

bhamiltoncx:
This implements setting the current working directory for a launched process on OS X, Linux, and Win32. Since calling `chdir()` affects every thread in the process, we have to use different strategies on each OS.

1. Win32: Pretty easy: `CreateProcessW()` already supports passing in the current directory as a `U+0000`-terminated array.
2. OS X: Also fairly easy: `syscall(SYS___pthread_chdir, cwd)` allows us to temporarily override the thread's current directory before we spawn the process, then we restore it.
3. Linux: This one is more difficult. We create a thread pool from which we use the `unshare(CLONE_FS)` API to disassociate the calling thread's cwd from the rest of the process. Then we call `chdir()` from that thread before we spawn the process.

I included a unit test. I made sure everything works on Windows, Mac, and Linux.

This does depend on the Java 7 `Path` API.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -34,6 +34,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.Future;
@@ -55,7 +56,7 @@ public abstract class BasePosixProcess implements NuProcess
    protected static final IEventProcessor<? extends BasePosixProcess>[] processors;
    protected static int processorRoundRobin;
 
-   private static ExecutorService linuxCwdChangeableExecutorService;
+   private static ExecutorService linuxCwdExecutorService;
 
    protected IEventProcessor<? super BasePosixProcess> myProcessor;
    protected volatile NuProcessHandler processHandler;
@@ -93,7 +94,7 @@ public abstract class BasePosixProcess implements NuProcess
    private Pointer posix_spawn_file_actions;
 
    // Launches threads under which changes to cwd do not affect the cwd of the process.
-   public static class LinuxCwdChangeableThreadFactory implements ThreadFactory {
+   public static class LinuxCwdThreadFactory implements ThreadFactory {
       private final AtomicInteger threadCount = new AtomicInteger(0);
 
       @Override
@@ -101,7 +102,11 @@ public abstract class BasePosixProcess implements NuProcess
          Runnable unshareCwdThenSpawn = new Runnable() {
             @Override
             public void run() {
+              // This makes a copy of the process's chroot, cwd, and umask and
+              // allows the thread to change any of those in a way that doesn't
+              // affect the rest of the process.
               int rc = LinuxLibC.unshare(LinuxLibC.CLONE_FS);
+              // If this throws, then it bubbles up to whomever called ExecutorService.submit().
               checkReturnCode(rc, "unshare(CLONE_FS) failed");
               r.run();
             }
@@ -144,12 +149,12 @@ public abstract class BasePosixProcess implements NuProcess
       }
 
       if (IS_LINUX) {
-         linuxCwdChangeableExecutorService = new ThreadPoolExecutor(
+         linuxCwdExecutorService = new ThreadPoolExecutor(
              /* corePoolSize */ numThreads,
              /* maximumPoolSize */ numThreads,
              /* keepAliveTime */ 0L, TimeUnit.MILLISECONDS,
              /* workQueue */ new LinkedBlockingQueue<Runnable>(),
-             /* threadFactory */ new LinuxCwdChangeableThreadFactory(),
+             /* threadFactory */ new LinuxCwdThreadFactory(),
              /* handler */ new ThreadPoolExecutor.DiscardPolicy());
       }
    }
@@ -295,9 +300,9 @@ public abstract class BasePosixProcess implements NuProcess
          StringArray environmentArray = new StringArray(environment);
          if (cwd != null) {
              if (IS_MAC) {
-                 rc = spawnWithCwdSetViaPthreadChdir(restrict_pid, commands[0], posix_spawn_file_actions, posix_spawnattr, commandsArray, environmentArray, cwd);
+                 rc = spawnOsxWithCwd(restrict_pid, commands[0], posix_spawn_file_actions, posix_spawnattr, commandsArray, environmentArray, cwd);
              } else if (IS_LINUX) {
-                 rc = spawnWithCwdSetViaLinuxCwdChangeableExecutorService(restrict_pid, commands[0], posix_spawn_file_actions, posix_spawnattr, commandsArray, environmentArray, cwd);
+                 rc = spawnLinuxWithCwd(restrict_pid, commands[0], posix_spawn_file_actions, posix_spawnattr, commandsArray, environmentArray, cwd);
              } else {
                  throw new RuntimeException("Platform does not support per-thread cwd override");
              }
@@ -371,7 +376,7 @@ public abstract class BasePosixProcess implements NuProcess
       return this;
    }
 
-   private int spawnWithCwdSetViaPthreadChdir(
+   private int spawnOsxWithCwd(
       IntByReference restrict_pid,
       String restrict_path,
       Pointer file_actions,
@@ -397,7 +402,7 @@ public abstract class BasePosixProcess implements NuProcess
      }
    }
 
-   private int spawnWithCwdSetViaLinuxCwdChangeableExecutorService(
+   private int spawnLinuxWithCwd(
       final IntByReference restrict_pid,
       final String restrict_path,
       final Pointer file_actions,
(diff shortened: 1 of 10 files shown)


src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -47,6 +55,8 @@
    protected static IEventProcessor<? extends BasePosixProcess>[] processors;
    protected static int processorRoundRobin;
 
+   private static ExecutorService linuxCwdChangeableExecutorService;
```
brettwooldridge: Shorten to ``linuxCwdExecutorService``.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -82,6 +92,26 @@
 
    private Pointer posix_spawn_file_actions;
 
+   // Launches threads under which changes to cwd do not affect the cwd of the process.
+   public static class LinuxCwdChangeableThreadFactory implements ThreadFactory {
```
brettwooldridge: Shorten to ``LinuxCwdThreadFactory``.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -82,6 +92,26 @@
...
+      private final AtomicInteger threadCount = new AtomicInteger(0);
+
+      @Override
+      public Thread newThread(final Runnable r) {
+         Runnable unshareCwdThenSpawn = new Runnable() {
+            @Override
+            public void run() {
+              int rc = LinuxLibC.unshare(LinuxLibC.CLONE_FS);
+              checkReturnCode(rc, "unshare(CLONE_FS) failed");
```
brettwooldridge: Who is going to catch the exception?

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -317,6 +369,62 @@ else if (WIFSIGNALED(status)) {
       return this;
    }
 
+   private int spawnWithCwdSetViaPthreadChdir(
```
brettwooldridge: The code is self-explanatory, so the method does not need to be so descriptive.  ``spawnOsxWithCwd()``.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -317,6 +369,62 @@ else if (WIFSIGNALED(status)) {
...
+     try {
+       return LibC.posix_spawnp(restrict_pid, restrict_path, file_actions, restrict_attrp, argv, envp);
+     } finally {
+       rc = LibC.SYSCALL.syscall(LibC.SYSCALL.SYS___pthread_chdir, oldCwd);
+       Native.free(Pointer.nativeValue(oldCwd));
+       checkReturnCode(rc, "syscall(SYS__pthread_chdir) failed to restore current directory");
+     }
+   }
+
+   private int spawnWithCwdSetViaLinuxCwdChangeableExecutorService(
```
brettwooldridge: The code is self-explanatory, so the method does not need to be so descriptive.  ``spawnLinuxWithCwd()``.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -317,6 +369,62 @@ else if (WIFSIGNALED(status)) {
...
+      final Pointer /*String[]*/envp,
+      final Path cwd)
+   {
+     Future<Integer> setCwdThenSpawnFuture = linuxCwdChangeableExecutorService.submit(
+         new Callable<Integer>() {
+           @Override
+           public Integer call() {
+               // Set cwd in this thread, which has its cwd state disassociated from the rest of the process.
+               int rc = LibC.chdir(cwd.toAbsolutePath().toString());
+               checkReturnCode(rc, "chdir() failed");
```
brettwooldridge: Who is going to catch the exception?  I think the Future is capable of reporting an exception...?

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -242,8 +243,14 @@ NuProcess start(List<String> commands, String[] environment)
          processInfo = new PROCESS_INFORMATION();
 
          DWORD dwCreationFlags = new DWORD(WinNT.CREATE_NO_WINDOW | WinNT.CREATE_UNICODE_ENVIRONMENT | WinNT.CREATE_SUSPENDED);
+         char[] cwdChars;
+         if (cwd != null) {
+           cwdChars = Native.toCharArray(cwd.toAbsolutePath().toString());
+         } else {
+           cwdChars = null;
+         }
```
brettwooldridge: Ternary.
```java
char[] cwdChars = (cwd != null) ? Native.toCharArray(cwd.toAbsolutePath().toString()) : null;
```

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -82,6 +92,26 @@
...
+   // Launches threads under which changes to cwd do not affect the cwd of the process.
+   public static class LinuxCwdChangeableThreadFactory implements ThreadFactory {
+      private final AtomicInteger threadCount = new AtomicInteger(0);
+
+      @Override
+      public Thread newThread(final Runnable r) {
+         Runnable unshareCwdThenSpawn = new Runnable() {
+            @Override
+            public void run() {
+              int rc = LinuxLibC.unshare(LinuxLibC.CLONE_FS);
```
brettwooldridge: One concern about ``unshare()``.  The doc says:
```
CLONE_FS
Unshare filesystem attributes, so that the calling process no longer
shares its root directory (chroot(2)), current directory (chdir(2)), or 
umask (umask(2)) attributes with any other process.
```
Does this allow the spawned process to escape a [chroot jail](http://docs.oracle.com/cd/E37670_01/E36387/html/ol_cj_sec.html)?

bhamiltoncx: Will clean up, thanks!

unshare(CLONE_FS) makes a copy of the existing settings including the
chroot, so it doesn't introduce a security flaw. (It would be a massive
security flaw if it changed the chroot!)

I'll document who has to catch the exceptions.

2015年9月5日(土) 0:10 Brett Wooldridge <notifications@github.com>:

brettwooldridge: Makes since on the ``unshare()``.

With respect to the exception, I think it is going to be swallowed by the executor, so there needs to be a way to propagate it out so that it is eventually thrown from the ``start()`` method.

bhamiltoncx: If I recall correctly, exceptions thrown inside a Runnable are caught by
the executor, then wrapped in an ExecutionException and thrown on the other
side. We'll want to unwrap them and re-throw them.
2015年9月5日(土) 7:41 Brett Wooldridge <notifications@github.com>:

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: That's a great question. If `ThreadFactory.newThread()` throws, it bubbles up to whomever called `ThreadPoolExecutor.execute()`.

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/concurrent/ThreadPoolExecutor.java#ThreadPoolExecutor.addWorkerFailed%28java.util.concurrent.ThreadPoolExecutor.Worker%29

I'd expect this to basically be a fatal. This system call shouldn't fail unless there's some serious kernel corruption going on.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: OK, done.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: The `RuntimeException` is caught then wrapped in an `ExecutionException` which is re-thrown when we call `Future.get()`.

Added a comment and updated to explicitly handle and unwrap `ExecutionException` thrown by `Future.get()`.

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Yeah, I confirmed this is the behavior.

brettwooldridge:
If you rebase this against the head, I'll merge it.

bhamiltoncx:
OK, rebased.

PR #34, opened 2015-09-08
Remove timeout in kevent call

bhamiltoncx:
Unlike Linux and Win32, OS X has async notifications of child process exit via `NOTE_EXIT`, so there's no point in waking up each processor thread every 100 ms (or whatever the linger and poll intervals indicate) and do nothing.

This diff simply removes the timeout passed to `kevent()`, since we don't do any work if there are no events raised anyway.


bhamiltoncx:
Oh, I see `BaseEventProcessor.run()` relies on this polling behavior. Let me make sure we can still shut down idle threads properly.

brettwooldridge:
Let me know when you think its good to go...

bhamiltoncx:
OK, fixed it up. We now pass a timeout iff `pidToProcessMap` is empty so we can linger up to `LINGER_TIME_MS` on OS X just like on Linux.

PR #37, opened 2015-09-10
Remove linux process reap and let BaseEventProcessor do its job

jamestyrrell:
@brettwooldridge We are still experiencing the same issue. I have dived into the code a bit more and it looks like that by exiting early the pipes are never written as the process is never registered. I am removed the early exit bit and now `BaseEventProcessor` is able to do its job.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -29,22 +29,9 @@ import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.CyclicBarrier;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@jamestyrrell What is your confidence level on this?  Have you really beaten on this change?

brettwooldridge:
Is the pull request obsoleted by pull #40?

jamestyrrell:
It will be, but I am not sure that catching an early exit is the best course of action. Why not just let `BaseEventProcessor` handle it?

brettwooldridge:
@jamestyrrell You can take another shot at it.

For certain Linux spawn failures, the [rc value](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java#L322) will be 0, and the [pid value](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java#L322) possibility 0 or undeterminable.

Removing the existing logic will result in ``registerProcess()`` getting called, which is likely to [encounter an issue here](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java#L79) because it relies on the ``pid``.  The ``cleanupProcess()`` and ``checkDeadPool()`` methods will similarly encounter issues [here](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java#L222),  [here](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java#L222), and [here](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java#L275).

Finally, if the ``NuProcessBuilder.start()`` method returns without exception, the user may call ``waitFor()`` or other methods that expect a valid ``pid``, and there is nothing we can do to protect them currently.

brettwooldridge:
The last issue above, that of returning from ``start()`` without exception, might cause issues if the user calls ``wantWrite()`` or ``stdinWrite()`` on the non-spawned process.  You would need to test what happens in those cases.

PR #42, opened 2015-09-15
Add "soft" close support to NuProcess.closeStdin()

bhamiltoncx:
Previously, there was no good way to write data to `stdin` of a `NuProcess` then close it once all the writes completed.

`CatTest` and a few examples use the assumption that the same number of bytes will be written to `stdout`, but that's not true for all processes the user might want to execute.

This diff changes `NuProcess.closeStdin()` to accept a `boolean force` parameter. If `true`, we use the old behavior of forcibly closing stdin and dropping all data pending write. If `false`, we enqueue a "tombstone" buffer to `pendingWrites` to indicate we should close stdin once that position in the queue is reached.

Once I did this, I found a few places where we were assuming `NuProcess.getStdin().get()` was always returning a valid fd. Since it's more common for it to return -1 now, I hardened the places where we assumed it did not. (This is still a little racy: it's possible to call `NuProcess.getStdin().get()` on one thread, then afterwards call `NuProcess.getStdin().set(-1); LibC.close(oldFd);` on another.)

Finally, I added a minor perf improvement to `BasePosixProcess`: after invoking `NuProcessHandler.onStdinReady()`, we'll immediately try to write the bytes to the process's `stdin` fd if there's any availability instead of waiting until the next time `kqueue()` / `epoll()` fires the stdin ready notification. (I think we can't currently make this improvement on Windows, since we only have one outstanding overlapped I/O at a time there.)

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -56,6 +56,7 @@ public abstract class BasePosixProcess implements NuProcess
    protected static final boolean IS_LINUX = System.getProperty("os.name").toLowerCase().contains("linux");
    private static final boolean LINUX_USE_VFORK = Boolean.parseBoolean(System.getProperty("com.zaxxer.nuprocess.linuxUseVfork", "true"));
    private static final boolean IS_SOFTEXIT_DETECTION;
+   private static final ByteBuffer STDIN_CLOSED_PENDING_WRITE_TOMBSTONE = ByteBuffer.allocate(1);
 
    protected static final IEventProcessor<? extends BasePosixProcess>[] processors;
    protected static int processorRoundRobin;
@@ -89,6 +90,7 @@ public abstract class BasePosixProcess implements NuProcess
    protected volatile int stdoutWidow;
    protected volatile int stderrWidow;
 
+   protected AtomicBoolean stdinClosing;
    protected boolean outClosed;
    protected boolean errClosed;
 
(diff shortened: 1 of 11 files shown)


src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -89,10 +89,12 @@
    protected volatile int stdoutWidow;
    protected volatile int stderrWidow;
 
+   protected AtomicBoolean stdinClosing;
    protected boolean outClosed;
    protected boolean errClosed;
 
    private ConcurrentLinkedQueue<ByteBuffer> pendingWrites;
+   private final ByteBuffer pendingWriteStdinClosedTombstone;
```
brettwooldridge: Make ``pendingWriteStdinClosedTombstone`` static final ``STDIN_CLOSED_PENDING_WRITE_TOMBSTONE``

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -118,19 +118,20 @@ public void queueWrite(LinuxProcess process)
 
       try {
          int stdin = process.getStdin().get();
+         if (stdin != -1) {
```
brettwooldridge: Prefer early return here.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
bhamiltoncx: Done.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
bhamiltoncx: Done.

bhamiltoncx:
Thanks, fixed those bits of feedback.

Issue #41, opened 2015-09-14
NuProcess start 1.0.3 fails on Windows 


clidev:
I'm unable to get 1.0.3 to work on Windows 7 (64-bit). It seems like NuProcess fails to create a process on Windows. The same test code works on Windows 7 with NuProcess 0.9.7.

Here's the stacktrace:
io.spikex.core.util.unit.ProcessTest > testCommandOutput STANDARD_ERROR
    java.lang.RuntimeException: CreateProcessW() failed, error: 2
        at com.zaxxer.nuprocess.windows.WindowsProcess.start(WindowsProcess.java:251)
        at com.zaxxer.nuprocess.windows.WinProcessFactory.createProcess(WinProcessFactory.java:42)
        at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:240)
        at io.spikex.core.util.process.ProcessExecutor.start(ProcessExecutor.java:115)

clidev:
I'm able to repeat the problem with the simple example code below. 

```java
package nuprocesstestwin7;

import com.zaxxer.nuprocess.NuProcess;
import com.zaxxer.nuprocess.NuProcessBuilder;
import java.util.concurrent.TimeUnit;
import java.util.logging.Level;
import java.util.logging.Logger;

public class NuProcessTestWin7 {

    public static void main(String[] args) {
        try {
            // Display JVM info
            System.out.println(System.getProperty("java.vendor")
                    + " " + System.getProperty("java.version"));
            // Display OS info
            System.out.println(System.getProperty("os.name")
                    + " " + System.getProperty("os.version")
                    + " (" + System.getProperty("os.arch") + ")");
            // Test output of DIR command
            NuProcessBuilder pb = new NuProcessBuilder("cmd.exe /c echo hi");
// This does not work either:
//            NuProcessBuilder pb = new NuProcessBuilder("cmd.exe", "/c", "echo hi");
            ProcessHandler handler = new ProcessHandler();
            pb.setProcessListener(handler);
            NuProcess process = pb.start();
            process.wantWrite();
            process.waitFor(0, TimeUnit.SECONDS); // when 0 is used for waitFor() the wait is infinite
        } catch (InterruptedException ex) {
            Logger.getLogger(NuProcessTestWin7.class.getName()).log(Level.SEVERE, null, ex);
        }
    }
}
```

This gives the following output:

```
Oracle Corporation 1.8.0_40
Windows 7 6.1 (amd64)
[...]

PR #43, opened 2015-09-17
Enable/disable command line quoting on Windows

clidev:
Added com.zaxxer.nuprocess.enableWindowsCmdArgQuoting system property to enable/disable command line argument quoting on Windows.

Also changed the pom to disable quoting by default so that the tests would pass on Windows.

Copied the non-quoting code from NuProcess 0.9.7. Not sure if the '\0' is required as terminating char for command lines, but it seemed that without it JNA Kernel32Test garbled the "echo" output. Better leave it then.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,6 +142,17 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


clidev:
This pull request tries to resolve issue #41.

bhamiltoncx:
https://github.com/brettwooldridge/NuProcess/pull/44 should fix the bug properly. Sorry about that.

brettwooldridge:
Fixed by pull request #44

PR #44, opened 2015-09-18
Correctly append space after each Win32 arg

bhamiltoncx:
I caused this regression in https://github.com/brettwooldridge/NuProcess/commit/4b8c94b60059f114713029bc9bd7f4a8b0b7bcac .

When I imported the escaping logic from Buck to NuProcess, I forgot that Buck's logic was outputting to a `List` to feed to `ProcessBuilder`, where `NuProcess` was outputting to a `StringBuilder`.

Since we're writing directly to a `StringBuilder`, we need to ensure we separate all the arguments by spaces.

I confirmed this fix by running `mvn test` on Windows.

This fixes https://github.com/brettwooldridge/NuProcess/issues/41 .


brettwooldridge:
No worries, happens to us all.  My bad for not having better unit test coverage.

PR #46, opened 2015-09-21
Fix FastExitingProcessTest on Windows

bhamiltoncx:
I found `FastExitingProcessTest` was failing on Windows. This is because `echo` is a `cmd.exe` built-in and literally echoes the double quotes around its args.

I worked around this by choosing a string which doesn't need to be escaped on Windows.

### DIFF
--- a/src/test/java/com/zaxxer/nuprocess/FastExitingProcessTest.java
+++ b/src/test/java/com/zaxxer/nuprocess/FastExitingProcessTest.java
@@ -64,15 +64,18 @@ public class FastExitingProcessTest
       Handler handler = new Handler();
       NuProcess process;
       if (Platform.isWindows()) {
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@bhamiltoncx Does this indicate an ongoing issue with the escaping logic?  I would expect NuProcess to detect that "Hello world!" contains a space, and therefore needs to be escaped automatically.  Admittedly there are various hand-wavy bits about whether it is already escaped/quoted, and the simple presence of a space may not need to trigger escaping.

brettwooldridge:
@bhamiltoncx Oh sorry, I read the code and not the comment.  In the case of echo, work to pass:
```java
process = new NuProcessBuilder(handler, "cmd.exe", "/c", "echo Hello world!").start();
```
?

PR #47, opened 2015-09-22
Remove UnsafeHelper, use ByteBuffer.allocateDirect() instead

bhamiltoncx:
I got the willies every time I built and saw the compiler warnings about `sun.misc.Unsafe` being deprecated.

Since JNA supports directly passing `ByteBuffer`s to methods which take `void *`, this diff removes `UnsafeHelper` in favor of directly using the process's `ByteBuffer`s. (This will also be a minor perf improvement, since we no longer need to allocate a new `Pointer` each time we call `read()` or `write()`, as we're always re-using the same `ByteBuffer` over and over).

Two things to note:

1. The native memory for the `ByteBuffer`s will now be freed on the background Java finalizer thread, instead of synchronously in `onExit()`.
2. We now rely on `inBuffer.hasRemaining()` to know whether there's any stdin data pending write. This means we need to set `inBuffer.limit(0)` when we first create it, to indicate there's no stdin data pending write.

I tested this on OS X and on Windows.


src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -616,10 +611,10 @@ public boolean writeStdin(int availability)
          return false;
       }
 
-      if (remainingWrite > 0) {
+      if (inBuffer.hasRemaining()) {
          int wrote;
          do {
-            wrote = LibC.write(fd, inBufferPointer.share(writeOffset), Math.min(remainingWrite, availability));
+            wrote = LibC.write(fd, inBuffer, Math.min(availability, inBuffer.remaining()));
```
brettwooldridge: I do have a question about this change.  How does LibC.write() know what offset to start the write from?  Does JNA automatically calculate the offset based on the fact that ``inBuffer`` is a ByteBuffer?

bhamiltoncx: Yes, JNA automatically obeys ByteBuffer.position() when marshaling
arguments. Here's their test to confirm that:

https://github.com/java-native-access/jna/blob/6d0ff952f872dc404a0b20067b74c9c4cde898ba/test/com/sun/jna/BufferArgumentsMarshalTest.java#L102
2015年9月22日(火) 18:16 Brett Wooldridge <notifications@github.com>:

bhamiltoncx: And here's the actual implementation:

https://github.com/java-native-access/jna/blob/c9383c5ce248aedb7eb201220a7c32700cac0451/native/dispatch.c#L1225

Ben
2015年9月22日(火) 18:16 Brett Wooldridge <notifications@github.com>:

brettwooldridge: Great, thanks.

PR #49, opened 2015-09-28
Fix wrong position and limit value on inBuffer caused by incorrect inBuffer clear.

pfxuan:
This bug happens when these three conditions happen together: no inBuffer remaining, no pending writes and no user wantsWrite. It results loading of whole empty inBuffer from position (0) to limit (BUFFER_CAPACITY).

### DIFF
--- a/src/test/java/com/zaxxer/nuprocess/DirectWriteTest.java
+++ b/src/test/java/com/zaxxer/nuprocess/DirectWriteTest.java
@@ -51,7 +51,7 @@ public class DirectWriteTest
    }
 
    // TODO: DirectWriteBig will explore a bug when using writeStdin at onStart()
-   @Test
+   //@Test
    public void testDirectWriteBig() throws InterruptedException
    {
       ProcessHandler2 processListener = new ProcessHandler2();
@@ -106,7 +106,9 @@ public class DirectWriteTest
 
       NuProcessBuilder pb = new NuProcessBuilder(processListener, command);
       NuProcess nuProcess = pb.start();
-      for (int i = 0; i < 10000; i++) {
+      // TODO: given a large i (e.g. 1,000, 10,000), this unit test (testConsecutiveWrites) will
+      //       produce a side-effect on InterruptTest (Mac OS X).
+      for (int i = 0; i < 100; i++) {
          ByteBuffer buffer = ByteBuffer.allocate(64);
          buffer.put("This is a test".getBytes());
          buffer.flip();
@@ -117,7 +119,7 @@ public class DirectWriteTest
 
       nuProcess.closeStdin(true);
       nuProcess.waitFor(0, TimeUnit.SECONDS);
-      Assert.assertEquals("Count did not match", 140000, count.get());
+      Assert.assertEquals("Count did not match", 1400, count.get());
    }
 
    private static class ProcessHandler1 extends NuAbstractProcessHandler
--- a/src/test/java/com/zaxxer/nuprocess/DirectWriteTest.java
+++ b/src/test/java/com/zaxxer/nuprocess/DirectWriteTest.java
@@ -51,6 +51,7 @@ public class DirectWriteTest
    }
 
    // TODO: DirectWriteBig will explore a bug when using writeStdin at onStart()
+   //       (has problem on Mac OS X and Linux, but works on Win32)
    //@Test
    public void testDirectWriteBig() throws InterruptedException
    {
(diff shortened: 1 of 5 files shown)


brettwooldridge:
@bhamiltoncx Care to review, as this impacts some of your recent changes?

bhamiltoncx:
Oops! Thank you. Two things:

1. I think the same bug exists in the win32 code, can you fix it?
2. A unit test would be very nice.

pfxuan:
@bhamiltoncx Got it! I'm going to write the unit test to reproduce this bug, and will patch win32 implementation later. Need to find a Windows machine :)

pfxuan:
@bhamiltoncx The unit test for exploring this issue already exists in code base. I did a few minor twists on `DirectWriteTest()` and make it work with new changes. It looks like current implementation has an issue for supporting consecutive direct writes on Mac OS platform. In addition, I found two new bugs after activating `DirectWriteTest()` unit test:   
1. testDirectWriteBig() does not work on Mac OS and Linux. So I temporary mute this test. 
2. testConsecutiveWrites() has a side-effect on InterruptTest() when using a large number of consecutive writes (e.g. 1,000, 10,000). This only happens on Mac OS X, but has not problem on Linux and Win32.

I'm not sure what causes these two issues, but left comments on TODO list. Please check commit for the details. Thanks!

src/test/java/com/zaxxer/nuprocess/DirectWriteTest.java
```diff
@@ -105,7 +107,9 @@ public void onStdout(ByteBuffer buffer, boolean closed)
 
       NuProcessBuilder pb = new NuProcessBuilder(processListener, command);
       NuProcess nuProcess = pb.start();
-      for (int i = 0; i < 1000; i++) {
+      // TODO: given a large i (e.g. 1,000, 10,000), this unit test (testConsecutiveWrites) will
+      //       produce a side-effect on InterruptTest (has problem on Mac OS X, but works on Linux and Win32).
+      for (int i = 0; i < 100; i++) {
```
brettwooldridge: We may be able to avoid the side-effect by [configuring the surefire](https://github.com/brettwooldridge/HikariCP/blob/dev/pom.xml#L294) plugin in this project's pom.xml and setting it to not reuse forks (``forkCount=1/reuseForks=true`` documentation [here](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html)).  

Worth a try.

pfxuan: Just did a quick test. This does the work.

brettwooldridge: That said, *something* is indeed up with the interrupting processes on OS X.  One out of every few runs of InterruptTest on OS X leaves one, two, or three zombie ``/bin/cat`` processes with *no parent* pid.  These processes are unkillable, even by root with ``kill -9``, the only recourse to get rid of them is reboot.

brettwooldridge: If the ``reuseForks=false`` does address the issue, I would say commit the pom.xml and bump the iterations here back up to 1000.

pfxuan: I did a bit troubleshooting on `InterruptTest` and found a dead loop at:  
```
      while (true) {
         if (count.get() > 10000) {
            process.destroy(forceKill);
            break;
         }
         Thread.sleep(20);
      }
```

brettwooldridge: Yeah, that can't be right.  Should be ``count.getAndIncrement()``.

pfxuan: `count.getAndIncrement() > 10, 000` is too long (20 x 10,000 = 200 seconds). Can we use >100 (2 seconds) for destroying process?

brettwooldridge: Sure.

pfxuan: So we should keep reusing forks on surefire. Right?

brettwooldridge: Yes.  We can sort out what is wrong with InterruptTest separately.

pfxuan: Okay. I'm going to fix `InterruptTest ` bug fist and comment the configuration on surefire for now.

pfxuan:
@brettwooldridge so far, everything works except `testDirectWriteBig()` on Mac OS X.

brettwooldridge:
Thanks for the hard work!

pfxuan:
You are welcome! This is a very useful and also promising project!

brettwooldridge:
@pfxuan @balonus Version 1.1.0 has been released, including a fix for the direct write test on Mac OS X.

pfxuan:
@brettwooldridge Wonderful! Thanks for letting me know.

PR #50, opened 2015-10-07
Signalfd

brettwooldridge:
Based on [that](https://ldpreload.com/blog/signalfd-is-useless), is this pull still worth it?


xkr47:
Is there a risk that processes spawned through traditional means e.g. java.lang.ProcessBuilder don't get their signals properly reset to defaults on POSIX systems?

PR #61, opened 2016-06-19
Add getPID() to NuProcess.

MTyson:
Added the getPID() method to NuProcess interface.

Implemented it in WindowsProcess and BasePosixProcess.

getPID() returns an integer representing the actual OS process id.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/NuProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/NuProcess.java
@@ -165,5 +165,5 @@ public interface NuProcess
     */
    void setProcessHandler(NuProcessHandler processHandler);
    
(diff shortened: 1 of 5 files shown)


src/main/java/com/zaxxer/nuprocess/NuProcess.java
```diff
@@ -164,4 +164,6 @@
     * @param processHandler the new {@link NuProcessHandler}
     */
    void setProcessHandler(NuProcessHandler processHandler);
+   
+   public int getPID();
```
brettwooldridge: Remove ``public`` keyword, all interface methods are public by default.

MTyson:
Done deal.

Issue #63, opened 2016-07-03
WindowsProcess printStackTrace


benor1470:
Hey, the WindowsProcess  uses printStackTrace (lines 173,220,249,303,332,392) and not using logger, can you please use a logger so i can turn in off when they aren't needed?
My log shows those errors on tests and without a way to turn them off they fill my log

Thank you

brettwooldridge:
Pull requests welcomed.

fanf:
I have the same problem with ```BasePosixProcess```. 
I have two concerns: 
- I can't adapt (and potentially recover at the application logic level) from the error that appears in the stacktrace, and which is not always an exception from the application logic
- I have non-contextualise stacktrace. Typically, when I get "Invocation of posix_spawn() failed, return code: 13, last error: 13", I would prefer to tell my user to look at the command permission (with the path to the command), but never ever display that message to them. 

To not depend to a given log framework, and idea could be to add a ```PrinStream``` to the process builder, and use that one in place of ```System.err``` if provided. 
Users who care would be able to provide a byte array output stream to post process the stream in case of exit code ```Int.MinValue```, and chose what to actually disaply to users using their application logging framework. 

Do you think the idea can fly ?

bturner:
In the years since this was last updated, `BasePosixProcess` has already been changed to replace `e.printStackTrace` with use of the JDK's `java.util.logging.Logger`, and 1.2.x and 2.0.x releases include that change. I've recently opened pull request #118 which applies a similar treatment to `WindowsProcess`. Hopefully that can be merged/released soon.

bturner:
My changes from pull request #118 were included in NuProcess 2.0.2, eliminating the last `e.printStackTrace` calls from the codebase.

PR #64, opened 2016-07-03
Added usage for slf4j

benor1470:
Added slf4j as a fix for issue #63

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,7 +115,7 @@
                        <Bundle-Name>NuProcess</Bundle-Name>
                        <Export-Package>com.zaxxer.nuprocess</Export-Package>
                        <Export-Package>com.zaxxer.nuprocess.codec</Export-Package>
-                       <Import-Package>com.sun.*,sun.misc.*</Import-Package>
+                       <Import-Package>com.sun.*,sun.misc.*,org.slf4j</Import-Package>
                        <Bundle-SymbolicName>${project.groupId}.${project.artifactId}</Bundle-SymbolicName>
                    </instructions>
                </configuration>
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@bhamiltoncx Does a dependency on slf4j cause Buck any grief?  Is java.util.logging better (no external dependency), allowing users to use jul-to-slf4j.jar if they so choose?

benor1470:
SLF4J provides much simple and powerful APIs when compared to Java Util Logger and it is a standard

brettwooldridge:
@benor1470 I understand it is widely used, I use it in many projects.  The issue is more of not breaking existing users who may not have a slf4j dependency.

bhamiltoncx:
Buck doesn't really want slf4j as a dependency. We use java.util.logging
with a thin shim (the same as Airlift) and it works fine.

benor1470:
Then I removed slf4j and used native java

pom.xml
```diff
@@ -110,7 +110,7 @@
                        <Bundle-Name>NuProcess</Bundle-Name>
                        <Export-Package>com.zaxxer.nuprocess</Export-Package>
                        <Export-Package>com.zaxxer.nuprocess.codec</Export-Package>
-                       <Import-Package>com.sun.*,sun.misc.*</Import-Package>
+                       <Import-Package>com.sun.*,sun.misc.*,org.slf4j</Import-Package>
```
brettwooldridge: This change seems unnecessary.  I am ready to accept this pull request after this is removed.

analytically:
There's always `jul-to-slf4j`: 

The jul-to-slf4j module includes a java.util.logging (jul) handler, namely SLF4JBridgeHandler, which routes all incoming jul records to the SLF4j API. 

http://www.slf4j.org/legacy.html

bturner:
It's nice to see this getting tweaked for `WindowsProcess`, but shouldn't the same change be made to the `printStackTrace` calls in `BasePosixProcess` and `ProcessCompletions` (which is also in the Windows-specific code) and others? This seems like a pretty incomplete change.

brettwooldridge:
Static loggers in a library is an [anti-pattern](http://wiki.apache.org/commons/Logging/StaticLog).

PR #65, opened 2016-08-30
Correct bundle Export-Package specification in pom.xml

lis0x90:
You have a little mistake in osgi bundle export-package spec for maven-bundle-plugin


lis0x90:
Please review. Its very small change, but it very important for OSGi based platforms

PR #93, opened 2018-07-15
Use ALIGN_NONE only on x86-64.

bturner:
- Added EpollEvent.EpollEventPrototype.detectAlignment(), which uses the "os.arch" JVM property to detect x86-64 and set ALIGN_NONE
  - On any other architecture, ALIGN_GNUC is used
  - When running a 32-bit JVM on a 64-bit OS, ALIGN_GNUC will be used
- Updated EpollEventTest.testSize to set the expected size to 12 or 16 bytes depending on the architecture and/or pointer size

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
@@ -76,20 +76,14 @@ class EpollEvent
             epoll_data_t data; // User data variable
           } __EPOLL_PACKED;
 
-          On x86, __EPOLL_PACKED is:    #define __EPOLL_PACKED
-          On x86_64, __EPOLL_PACKED is: #define __EPOLL_PACKED __attribute__ ((__packed__))
-
-          sizeof(struct epoll_event) is 12 on x86 and x86_64
+          sizeof(struct epoll_event) is 12 on x86 and x86_64, but is 16 on other 64-bit platforms
       */
 
       public int events;
       public EpollData data;
 
       EpollEventPrototype() {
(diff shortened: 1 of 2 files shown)


bturner:
@brettwooldridge This is the line I'm thinking along.

One thing I was a bit concerned about was running 32-bit Java on a 64-bit OS, but I _think_ it should work correctly. If the JVM is a 32-bit process, it will use one of the 32-bit `libjnidispatch` libraries.

A different approach might be to use JNA's `Platform` to make choices. Since it has a fair amount of normalization built on experience; I've opened pull request #94 for that approach.

src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
```diff
@@ -106,6 +100,25 @@ int getFieldOffset(String field)
...
+       * #endif
+       * </pre></code>
+       *
+       * On x86-64 (amd64) platforms, {@code ALIGN_NONE} is used (to emulate {@code __attribute__((packed))}),
+       * and on all other platforms {@code ALIGN_GNUC} is used.
+       */
+      static int detectAlignment() {
+         String arch = System.getProperty("os.arch");
+
+         return "amd64".equals(arch) || "x86-64".equals(arch) ? ALIGN_NONE : ALIGN_GNUC;
```
bturner: Based on JNA's [getCanonicalArchitecture](https://github.com/java-native-access/jna/blob/master/src/com/sun/jna/Platform.java#L235-L260) this probably needs to test for "x86_64" too

bturner:
Declining in favor of #94, which has already been merged.

PR #94, opened 2018-07-15
Use ALIGN_NONE only on x86-64.

bturner:
- Added `EpollEvent.EpollEventPrototype.detectAlignment()`, which uses JNA's `Platform` utilities to detect x86-64 and use `ALIGN_NONE`
  - On any other architecture, `ALIGN_GNUC` is used
  - When running a 32-bit JVM on a 64-bit OS, `ALIGN_GNUC` will be used
- Updated `EpollEventTest.testSize` to set the expected size to 12 or 16 bytes depending on the architecture detected by JNA's `Platform`


bturner:
This is similar to #93, but it uses JNA's `Platform` to make decisions instead of trying to parse "os.arch" directly.

brettwooldridge:
@bturner I'll take this one. 👍

Issue #95, opened 2018-08-09
Pump on calling thread


bturner:
The way NuProcess reuses a fixed set of pump threads for handling process output is a great feature, and helps control thread bloat. It's one of the reasons we want to migrate Bitbucket Server over to it. However, as I make progress switching from Java's `ProcessBuilder` to NuProcess, I'm finding it's not always the pumping approach that we want.

When processing requests in Bitbucket Server, the most common scenario is that the system starts a `git` process and then blocks in `waitFor`, on that thread, to let it complete. During that time, that calling thread is doing nothing while one of the fixed pool of pump threads interleaves pumping for the started process, and most likely other processes, until the process completes. At that point, `waitFor` returns and I can let the calling thread continue.

What I'd like to propose is an alternate method, in addition to the current `NuProcessBuilder.start` that builds a `NuProcess` and, instead of calling `registerProcess` to assign it to one of the shared pump threads, (essentially) creates a dedicated `ProcessEpoll` (or equivalent) and runs the processing loop directly on the calling thread, only returning when the process completes (or some timeout expires).

_I'm willing to build this myself._ I'm not asking for someone else to dedicate their free time to enabling my use case. But if such a change has no chance of getting accepted, well then there's not much point.

### Some context on why I feel this is important

Bitbucket Server has essentially three "classes" of processes:

* Hosting processes (`git receive-pack`, `git upload-pack`, `git http-backend`)
* Command processes (all your "simple", _generally short-lived_ processes like `git diff`, `git log`, `git rev-list`)
* Custom processes, started by third-party plugins which are installed into Bitbucket Server
  * We provide a public API and SPI, exported via OSGi, which third parties are able to use to build their own custom integrations. Those integrations are installed _in process_ with the rest of Bitbucket Server
  * One of the things such integrations can do is use our API to spawn their own `git` (and other) processes, with whatever command line they need
  * We're not able to limit or control the _number_ of such processes third party integrations start

[...]

PR #62, opened 2016-06-22
Windows environment variables

ilya-klyuchnikov:
fixes #60 
Tests:
https://ci.appveyor.com/project/ilya-klyuchnikov/nuprocess/build/13 (before fix)
https://ci.appveyor.com/project/ilya-klyuchnikov/nuprocess/build/14 (after fix)

PR #103, opened 2019-08-31
Handle SystemRoot case-insensitively (Windows).

bturner:
- Updated `WindowsProcess.getEnvironment(String[])` to perform a case-insensitive lookup to detect when "SystemRoot" is already included in the supplied environment array
  - Without this check processes started by NuProcess can end up with both "SystemRoot" and "SYSTEMROOT" set
- Updated `EnvironmentTest` to use a `LinkedHashMap` to hold the environment from the `ProcessBuilder` and `NuProcess` processes, to retain the order the variables appear in
  - Removed System.out calls

This handling emulates `java.lang.ProcessEnvironment.toEnvironmentBlock`, from the Windows version of the JDK, which also uses a case-insensitive check to detect "SystemRoot" in the supplied environment and copies the value from the running application's environment if it's not set.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
@@ -44,6 +44,7 @@ public final class WindowsProcess implements NuProcess
    private static final boolean IS_SOFTEXIT_DETECTION;
 
    private static final int BUFFER_SIZE = 65536;
+   private static final String ENV_SYSTEMROOT = "SystemRoot";
 
    private static final ProcessCompletions[] processors;
    private static int processorRoundRobin;
@@ -90,7 +91,7 @@ public final class WindowsProcess implements NuProcess
          processors[i] = new ProcessCompletions();
       }
 
-      if (Boolean.valueOf(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
+      if (Boolean.parseBoolean(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
          Runtime.getRuntime().addShutdownHook(new Thread(new Runnable() {
             @Override
             public void run()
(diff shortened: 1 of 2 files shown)


bturner:
At the end of pull request #62, @ilya-klyuchnikov noted that the change as merged caused failures in the AppVeyor build. In the [build output](https://ci.appveyor.com/project/ilya-klyuchnikov/nuprocess/build/22), you can see that the `ProcessBuilder` process only has "SYSTEMROOT", but the NuProcess process has both "SYSTEMROOT" and "SystemRoot". When I build NuProcess on Windows, I get exactly that failure.

In his comment, @ilya-klyuchnikov linked to a commit which fixed the issue. That link is now dead (perhaps he deleted his fork?), but fortunately, with how Github handles fork hierarchies, it's actually possible to [view the commit](https://github.com/brettwooldridge/NuProcess/commit/b8f2f9fc78e82731271c28780935899e2b7a2880) using this repository instead (even though said commit isn't actually referenced in this repository). My changes here are essentially the changes he proposed then, with a few minor alterations.

I'm happy to make any adjustments, if something isn't right. Just let me know.

(For context, we're working on submitting a pull request for issue #95, as part of moving forward with adopting NuProcess for Bitbucket Server. As part of that work, we've set up internal CI to build our changes while we're developing them. However, the Windows job always fails due to this issue.)

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -90,7 +91,7 @@
          processors[i] = new ProcessCompletions();
       }
 
-      if (Boolean.valueOf(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
+      if (Boolean.parseBoolean(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
```
bturner: Sorry, meant to revert this.

src/test/java/com/zaxxer/nuprocess/EnvironmentTest.java
```diff
@@ -69,16 +68,13 @@ public void defaultEnv() throws InterruptedException, IOException
 		ProcessBuilder pb = new ProcessBuilder(command);
 		pb.environment().clear();
 		pb.environment().putAll(env);
-		System.out.println("Started Java Process");
-		Process process = pb.start();		
-		System.out.println("Waited for Java Process");
-		Set<String> result = new HashSet<String>();
```
bturner: Since these results are going to be compared using `assertEquals`, hash order vagaries concern me. I haven't seen any failures, but a more consistent ordering doesn't seem like a bad idea. Happy to revert, if you disagree (or to switch to `TreeSet` if you think we shouldn't care about the actual order, and only about what variables are set)

brettwooldridge:
Thanks for the contribution!

bturner:
Big thanks for the quick turnaround, @brettwooldridge!

PR #104, opened 2019-09-02
Adds ability to run processes synchronously (#95).

dikeert:
As per discussed in #95, this one implements ability to run processes synchronously with pumping happening on the client thread.

To achieve that, it extends builder, factory and process classes with `run` method that starts the process and then creates new event processor per client to listen for events. Processor then executes event loop the same way as it does for asynchronous code path.

The change is tested with new `runProcessSynchronously` test in `CatTest`.

Additionally it bumps minor version, because we've added a new feature here.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
+++ b/src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
@@ -253,16 +253,41 @@ public class NuProcessBuilder
     */
    public NuProcess start()
    {
+      ensureListener();
+      String[] env = prepareEnvironment();
+
+      return factory.createProcess(command, env, processListener, cwd);
+   }
+
+   /**
+    * Spawn the child process with the configured commands, environment, and {@link NuProcessHandler}
+    * and wait for it to complete running.
+    *
+    * @since 1.3
+    */
+   public void run()
+   {
+      ensureListener();
+      String[] env = prepareEnvironment();
+
+      factory.runProcess(command, env, processListener, cwd);
+   }
+
+   private void ensureListener()
+   {
       if (processListener == null) {
          throw new IllegalArgumentException("NuProcessHandler not specified");
       }
+   }
 
+   private String[] prepareEnvironment()
+   {
       String[] env = new String[environment.size()];
       int i = 0;
       for (Entry<String, String> entrySet : environment.entrySet()) {
          env[i++] = entrySet.getKey() + "=" + entrySet.getValue();
       }
 
-      return factory.createProcess(command, env, processListener, cwd);
+      return env;
    }
 }
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -44,6 +44,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
    private int epoll;
    private EpollEvent triggeredEvent;
    private List<LinuxProcess> deadPool;
+   private LinuxProcess process;
 
    static
    {
@@ -56,6 +57,23 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
 
    ProcessEpoll()
    {
+      this(LINGER_ITERATIONS);
+   }
+
+   ProcessEpoll(LinuxProcess process)
+   {
+      this(-1);
+
+      this.process = process;
+
+      registerProcess(process);
+      checkAndSetRunning();
+   }
+
+   private ProcessEpoll(int lingerIterations)
+   {
+      super(lingerIterations);
+
       epoll = LibEpoll.epoll_create(1024);
       if (epoll < 0) {
          throw new RuntimeException("Unable to create kqueue: " + Native.getLastError());
@@ -165,6 +183,19 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
    }
 
    @Override
+   public void run() {
+      super.run();
+
+      if (process != null) {
+         // For synchronous execution, perform a final _blocking_ deadpool check. If the deadpool
+         // is empty, this will return immediately. If the process has already exited, waitpid will
+         // return immediately. Otherwise, this will block until the the process terminates. This
+         // is necessary to ensure the handler's onExit is called before LinuxProcess.run returns.
+         checkDeadPool(0);
+      }
+   }
+
+   @Override
    public void closeStdin(LinuxProcess process)
    {
       try {
@@ -256,7 +287,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
                linuxProcess.getStderr().release();
             }
          }
-         checkDeadPool();
+         checkDeadPool(LibC.WNOHANG);
       }
    }
 
@@ -294,7 +325,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       }
    }
 
-   private void checkDeadPool()
+   private void checkDeadPool(int options)
    {
       if (deadPool.isEmpty()) {
          return;
@@ -304,7 +335,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       Iterator<LinuxProcess> iterator = deadPool.iterator();
       while (iterator.hasNext()) {
          LinuxProcess process = iterator.next();
-         int rc = LibC.waitpid(process.getPid(), ret, LibC.WNOHANG);
+         int rc = LibC.waitpid(process.getPid(), ret, options);
          if (rc == 0) {
             continue;
          }
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -183,7 +183,8 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
    }
 
    @Override
-   public void run() {
+   public void run()
+   {
       super.run();
 
       if (process != null) {
@@ -287,6 +288,17 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          }
          checkDeadPool();
       }
+   }
+
+   /**
+    * Closes the {@code eventpoll} file descriptor.
+    *
+    * @since 1.3
+    */
+   @Override
+   protected void close()
+   {
+      LibC.close(epoll);
    }
 
    // ************************************************************************
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -187,11 +187,9 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       super.run();
 
       if (process != null) {
-         // For synchronous execution, perform a final _blocking_ deadpool check. If the deadpool
-         // is empty, this will return immediately. If the process has already exited, waitpid will
-         // return immediately. Otherwise, this will block until the the process terminates. This
-         // is necessary to ensure the handler's onExit is called before LinuxProcess.run returns.
-         checkDeadPool(0);
+         // For synchronous execution, wait until the deadpool is drained. This is necessary to ensure
+         // the handler's onExit is called before LinuxProcess.run returns.
+         waitForDeadPool();
       }
    }
 
@@ -287,7 +285,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
                linuxProcess.getStderr().release();
             }
          }
-         checkDeadPool(LibC.WNOHANG);
+         checkDeadPool();
       }
    }
 
@@ -325,7 +323,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       }
    }
 
-   private void checkDeadPool(int options)
+   private void checkDeadPool()
    {
       if (deadPool.isEmpty()) {
          return;
@@ -335,7 +333,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       Iterator<LinuxProcess> iterator = deadPool.iterator();
       while (iterator.hasNext()) {
          LinuxProcess process = iterator.next();
-         int rc = LibC.waitpid(process.getPid(), ret, options);
+         int rc = LibC.waitpid(process.getPid(), ret, LibC.WNOHANG);
          if (rc == 0) {
             continue;
          }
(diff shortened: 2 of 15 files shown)


src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
```diff
@@ -252,17 +252,42 @@ public void setCwd(Path cwd)
...
+      return factory.createProcess(command, env, processListener, cwd);
+   }
+
+   /**
+    * Spawn the child process with the configured commands, environment, and {@link NuProcessHandler}
+    * and wait for it to complete running.
+    *
+    * @since 1.3
+    */
+   public void run()
```
dikeert: Me and @bturner have considered to return several things here:

- The `NuProcess` itself. This doesn't make sense as what for would you use it, given that the method returns after process is exited?
- The status code from syscalls that start processes. I struggle to imagine use case for this one.
- The exit code of the process. But in this case it would duplicate the code that is sent to `NuProcessHandler#onExit`.

We've chosen to return nothing but we really don't have strong opinion on that matter and ready to change it to your preferred way if you have one.

The same comment applies for whole chain of `run` methods in builder and factory.

bturner:
I've reviewed this change internally, prior to @dikeert opening this pull request, just to try and ensure we put up something as straightforward as possible.  We've tried pretty hard to keep the change as small as possible, avoiding incidental changes where possible. While this change still ends up being fairly large, it's adding a fairly big feature.

As @dikeert indicated, we're happy to make any changes you'd like to see in order to move this forward. As discussed on #95, getting some form of pump-on-thread synchronous execution added to NuProcess is critical to our ability to use it for Bitbucket Server. We recognize that your time is both valuable and limited, so if there's anything we can do here to help, please let us know!

(Also, note sure what's up with the Shippable build; that doesn't seem like it's related to anything we've changed here. That looks more like an infrastructure failure to me.)

bturner:
Apologies @brettwooldridge, this isn't quite ready for review. `CatTest` isn't quite rigorous enough in its checks to detect the case when `onStdinCharsReady` and `onStdoutChars` are both _never called_, so `runProcessSynchronously` passes even though, on Windows, the stdin/stdout/stderr are never actually pumped. That's a trivial fix, but once that's fixed there's some secondary issue with processes hanging.

If you haven't already reviewed, please hold off while we do some further testing, and add some additional unit tests.

bturner:
Just noting that I've found the issue(s) with the Windows changes and have some local fixups that (legitimately) pass all the tests. We'll circle back with an update to this pull request soon.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -164,6 +182,19 @@ public void queueWrite(LinuxProcess process)
...
+   @Override
+   public void run() {
+      super.run();
+
+      if (process != null) {
+         // For synchronous execution, perform a final _blocking_ deadpool check. If the deadpool
+         // is empty, this will return immediately. If the process has already exited, waitpid will
+         // return immediately. Otherwise, this will block until the the process terminates. This
+         // is necessary to ensure the handler's onExit is called before LinuxProcess.run returns.
+         checkDeadPool(0);
```
bturner: I don't love this solution, but I'm torn between different options without any that I think are great. I'm keen to get your feedback.

The problem: With async processing, after a given process terminates, the pump thread is still running for some number of linger iterations (or potentially even longer if there are other active processes), and it's periodically probing the `deadPool` for processes that have finished up. That means, eventually, `waitpid` is going to detect that the process ended and `onExit` will be called. For _synchronous_ processes, after this `run()` call returns, the `deadPool` will never be checked again. That means if the process hasn't _fully exited_ by this point, `onExit` will never happen (and the process will become a zombie, since nothing will ever wait for it).

Options I considered:
* Instead of using a -1 for `lingerIterations`, use 0. That effectively requires at least one iteration (since the check is `>` rather than `>=`). This works, but effectively adds a _500ms delay_ to _every synchronous process_ (250ms for an extra loop to check the `deadPool` a second time; this loop will reap the process, but won't be considered "idle", and another 250ms for the loop _after_ that one, to actually show as "idle")
  * That's a huge penalty, especially for something like Bitbucket Server that forks dozens or hundreds of concurrent processes and can fork out several very short-lived processes in rapid succession to complete a single task
* Add `protected` method to `BaseEventProcessor` and update its `run` loop to call that instead of, or in addition to, checking `pidToProcessMap.isEmpty()`, which this class could override to also check `deadPool.isEmpty()` (like `ProcessCompletions` does on Windows)
  * This could drop the penalty as low as 250ms, since I can pair it with `-1` for `lingerIterations`, but it still _has_ a penalty.
* Build a separate Java loop that does a much shorter sleep (5-10ms) and still uses `WNOHANG` with `waitpid`
  * This further shortens the penalty, at the expense of a bit more code
* Run a "tight loop" of `waitpid` calls with `WNOHANG` with no delay between them
  * This pays a lot of Java -> native -> Java overhead, but doesn't introduce artificial delays

The penalties of the different approaches are pretty obvious in the execution duration for the new `RunTest`:
* As implemented here on this pull request, on my machine the test (as in, all test cases) completes in ~800 _milliseconds_ (always less than 1 second)
* With `0` for iterations, it completes in 32 _seconds_
* With the new `protected` method, it completes in 18 _seconds_
* With the Java loop:
  * With a 10ms `Thread.sleep`, the test completes in 1.5 _seconds_
  * With a 50ms `Thread.sleep`, the test completes in 2.8 _seconds_
  * Ultimately, the penalty here will basically be whatever sleep we apply
* With the "tight loop", it completes in ~850ms
  * Repeated runs of this and the blocking `waitpid` on the pull request tend to show ~20-30ms is all that separates them; in practice, the penalty with the tight loop is pretty negligible

I picked this approach because it was the fastest and the simplest. However, it does come with the drawback that, if the process "never ends" (perhaps some sort of deadlock or other hang, which prevents it from ever terminating), this approach is basically "lost" forever waiting for it.

If we think that's a significant risk, I'd probably just go with the tight loop and pair that with some sort of either iteration count or maximum wait time where the loop "gives up". (One option for the "give up" path might be to belatedly register the process in the deadpool of one of the background processor threads, but the coordination for that is a bit complicated since the background threads are managed by the process factories rather than the processor types; e.g., `LinProcessFactory` manages the pool rather than `ProcessEpoll`)

We're happy to make any changes to use whatever approach you like best. I'm just trying to offer some details on the issue here and some insight into the reasoning for the change as included on the review.

src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
```diff
@@ -103,14 +87,87 @@ public NuProcess start(List<String> command, String[] environment, Path cwd) {
...
+      finally {
+         LibC.posix_spawnattr_destroy(posix_spawnattr);
+         LibC.posix_spawn_file_actions_destroy(posix_spawn_file_actions);
+
+         // After we've spawned, close the unused ends of our pipes (that were dup'd into the child process space)
+         closePipes();
+      }
+   }
+
+   private int prepareProcess(String[] environment, Path cwd, String[] commands,
```
bturner: This is the same logic that was previously in `OsxProcess.start`, refactored into a separate method for reuse in `run` but otherwise unchanged

src/main/java/com/zaxxer/nuprocess/windows/ProcessCompletions.java
```diff
@@ -69,6 +71,25 @@
...
+
+      if (process == null) {
+         throw new IllegalArgumentException("process");
+      }
+
+      registerProcess(process);
+      checkAndSetRunning();
+   }
+
+   private ProcessCompletions(int lingerIterations)
```
bturner: This constructor already existed on `BaseEventProcessor`, and the `ProcessEpoll` and `ProcessKqueue` subclasses, but we had to add it on `ProcessCompletions`. Without the ability to specify the `lingerIterations`, the current hard-coded value applies a 2.5 _second_ penalty for every synchronous process. (Unless you override it with a system property, of course, but at that point you're also reducing iterations for async processes.)

src/main/java/com/zaxxer/nuprocess/windows/ProcessCompletions.java
```diff
@@ -89,10 +110,13 @@
...
       try {
-         startBarrier.await();
+         // If the process is running synchronously, startBarrier will be null
+         if (startBarrier != null) {
+            startBarrier.await();
+         }
 
          int idleCount = 0;
-         while (!isRunning.compareAndSet(idleCount > LINGER_ITERATIONS && deadPool.isEmpty() && completionKeyToProcessMap.isEmpty(), false)) {
+         while (!isRunning.compareAndSet(idleCount > lingerIterations && completionKeyToProcessMap.isEmpty() && deadPool.isEmpty() && pendingPool.isEmpty(), false)) {
```
bturner: I've added the `pendingPool` into this check because otherwise for synchronous processes (where `lingerIterations` is `-1`) `process()` never gets called the first time and the process is never actually added to `completionKeyToProcessMap`.

This does apply a slight behavior change for async processes, but I don't think it's a bad one. With this change, if the background processor thread was about to run its _final idle iteration_ and a `registerProcess` call came in, the thread would _not_ idle out; instead, the process would be picked up and pumped.

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -270,6 +246,64 @@ NuProcess start(List<String> commands, String[] environment, Path cwd)
...
+         NuKernel32.CloseHandle(hStdoutWidow);
+         NuKernel32.CloseHandle(hStderrWidow);
+      }
+
+      // This must happen after the widow handles are closed; otherwise the process loop will
+      // never end because stderr and stdout will never be detected as closed
+      myProcessor.run();
+   }
+
+   private void prepareProcess(List<String> commands, String[] environment, Path cwd)
```
bturner: This is the same logic that was previously in `WindowsProcess.start`, refactored into a separate method for reuse in `run` but otherwise unchanged

src/test/java/com/zaxxer/nuprocess/CatTest.java
```diff
@@ -335,6 +337,7 @@ public void softCloseStdinAfterWrite() throws Exception
          this.writeAdler32 = new Adler32();
 
          bytes = getLotsOfBytes();
+         exitCode = -1;
```
bturner: Without this, the exit code on the handler defaults to `0`. That means if `onExit` is _never called_, the `getExitCode() == 0` checks in the various tests will still pass. By setting this to `-1` (which, I'll add, `Utf8DecodingListener` already does), if `onExit` is never called the tests will fail.

src/test/java/com/zaxxer/nuprocess/CatTest.java
```diff
@@ -382,9 +385,9 @@ int getExitCode()
 
       boolean checkAdlers()
       {
-         return readAdler32.getValue() == writeAdler32.getValue();
+         return writes == WRITES && readAdler32.getValue() == writeAdler32.getValue();
```
bturner: Without this, if `onStdinReady` is _never called_, the `readAdler32` and `writeAdler32` values will match, since they're both empty. By checking the `writes` count we can ensure that not only do the values match, but they match after processing the expected number of writes.

src/test/java/com/zaxxer/nuprocess/RunTest.java
```diff
@@ -0,0 +1,468 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.zip.Adler32;
+
+/**
+ * Performs <i>synchronous</i> tests using {@link NuProcessBuilder#run}.
+ */
+public class RunTest
```
bturner: This file is essentially a copy of `CatTest`, but it uses `run` instead of `start`. The goal here is to apply all the same validations that are applied for asynchronous execution to synchronous execution. (This test turned up the `onExit` issue on Linux, for example)

src/test/java/com/zaxxer/nuprocess/RunTest.java
```diff
@@ -0,0 +1,468 @@
...
+   public void setup()
+   {
+      command = "cat";
+      if (System.getProperty("os.name").toLowerCase().contains("win")) {
+         command = "src\\test\\java\\com\\zaxxer\\nuprocess\\cat.exe";
+      }
+   }
+
+   @Test
+   public void lotOfProcesses() throws InterruptedException
```
bturner: This test is a little different from `CatTest.lotOfProcesses()`. That test kicks off 100 processes, which get portioned across the background threads and then it waits for them all to complete. It does that 20 times. (2000 processes total)

This test spins up 2-4 daemon threads and each thread runs 20 processes _serially_ and then returns. (40-80 processes total) I could do 100 processes serially on each thread, but I wasn't sure there was a lot of value in the extra invocations.

bturner:
@brettwooldridge I think this pull request should now be ready for review. I've fixed all the issues I could find (though I'm keen to get your feedback on whether you'd like any of them solved differently), and significantly expanded the test coverage for synchronous processes.

I've also updated Bitbucket Server to build against a snapshot of these changes and run some of its own tests. I've verified that those tests all pass on Linux, macOS and Windows. (At this point we're working internally on switching more of our code over to NuProcess, which will steadily increase the amount of second-hand NuProcess coverage our tests provide.)

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
bturner: Re: "if the process "never ends" (perhaps some sort of deadlock or other hang, which prevents it from ever terminating)", it's worth bearing in mind that the process would have to hit that point after closing both `stdout` and `stderr`, to trigger the soft exit detection to add the process to the `deadPool`. If either of those is still open, then regardless of how we handle the `deadPool`, it won't matter because we won't get that far. This thread will still be in the `process()` loop in `run()`, waiting for the soft exit.

brettwooldridge:
I need to review this fully, but one possible recommendation for handling a process that never exits, and therefore hangs the runSynchronously() call, is to allow a maxRuntime parameter to be provided.  What do you think?

bturner:
An execution timeout would be a nice improvement. The tricky part is going to be deciding how to apply it, though. `waitpid` doesn't have any form of timeout. We either use `WNOHANG` and get zero wait at all (perhaps surrounded by some wait in Java), or we don't pass any options and it sits until it's done. I don't know whether a `Thread.interrupt()` in Java-land would interrupt the `waitpid` call, but I suspect it wouldn't.

Do you have any preferences on how such a timeout would be applied? Tight loop rechecking? Short sleep in Java to "slow down" the tight loop? (Bearing in mind that any sleep is going to basically be a straight runtime penalty for all well-behaved processes, as I wrote up above)

bturner:
What do you think about a Java wait with a backoff? Maybe 2 requests with no delay, then 1ms, then 3ms, then 7ms, etc? That way the penalty is minimal for well-behaved processes, but we don't hang indefinitely hammering away at `waitpid` for poorly-behaved ones?

bturner:
@brettwooldridge I've switched back to `WNOHANG` for checking synchronous termination, with a Java loop that applies a double-plus-one backoff between checks. `RunTest` appears to complete in approximately the same amount of time with this looping approach as it did with the blocking wait.

I added some `System.out.println` calls (removed before I committed) to verify that the loop _does_ sometimes start sleeping. My testing never showed it going past `Thread.sleep(1L)` before termination was detected, which seems good. I saw three different cases:
* One deadpool check, no sleep and termination
* Two deadpool checks, still no sleep and termination
* Three deadpool checks, with a 1ms sleep between the second and third check, and termination

If this approach doesn't work for you, or if you've got some improvements in mind, just let me know what you'd like to see.

bturner:
Thanks for the merge, and the 2.0.0 release, @brettwooldridge.

We've just shipped Bitbucket Server 7.0.0 which now uses NuProcess by default for (almost) all of its process management. We shipped with a fork of NuProcess due to timing, but I've already opened a pull request to switch over to the official 2.0.0 release for 7.1.

There are a few things like `@since` tags that are showing 1.3.0, in this change. Would you like me to raise a separate pull request to correct those things?

Issue #91, opened 2018-06-25
jdk-10.0.1/lib/libjava.so: undefined symbol: Java_java_lang_UNIXProcess_forkAndExec


fanf:
With NuProcess version 1.2.3, I get the following error when I try to run my tests with Open JDK 10.0.1 under linux: 

UNIXProcess_forkAndExec': /home/fanf/java/jdk-10.0.1/lib/libjava.so: undefined symbol: Java_java_lang_UNIXProcess_forkAndExec
        at com.sun.jna.Function.<init>(Function.java:245)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:566)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:542)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:528)
        at com.sun.jna.Native.register(Native.java:1777)
        at com.sun.jna.Native.register(Native.java:1370)
        at com.zaxxer.nuprocess.internal.LibJava8.<clinit>(LibJava8.java:36)
        at com.zaxxer.nuprocess.internal.BasePosixProcess.start(BasePosixProcess.java:163)
        at com.zaxxer.nuprocess.linux.LinProcessFactory.createProcess(LinProcessFactory.java:40)
        at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:266)
        at com.normation.rudder.hooks.RunNuCommand$.run(RunNuCommand.scala:153)

I didn't found other reference to problem with OpenJDK 10 (appart #87 but it seems different). It used to work with 1.1.x branch. 
Do you need more information to reproduce?

slandelle:
As of Java 10, UNIXProcess was merged into ProcessImpl: http://hg.openjdk.java.net/jdk10/jdk10/jdk/rev/98eb910c9a97

Neoware:
@brettwooldridge Hi ! If you can check my pull request :)

slandelle:
Basically, it seems that Azul has back-ported into their Java 8 build the change I mentioned above.
The PR:
* fixes `Constants#JVM_MAJOR_VERSION` that was broken with Java9+ and was actually unused
* renames the new stategy that was introduced for Azul as it's actually generic for Java10+
* uses this strategy when running on Azul or Java10+

fanf:
@brettwooldridge @slandelle @Neoware seems great (but I know nothing about nuprocess internals)

[...]

Issue #87, opened 2018-05-09
Upgrading to 1.2.x (from 1.1.3) causes NoClassDefFoundError in macOS (and Java 10)


robby-phd:
Upgrading to 1.2.*x* throws the following exception when using NuProcess in macOS (10.13.4, Zulu JDK 10.2+3-jdk10.0.1):

```
java.lang.NoClassDefFoundError: Could not initialize class com.zaxxer.nuprocess.osx.OsxProcess
	at com.zaxxer.nuprocess.osx.OsxProcessFactory.createProcess(OsxProcessFactory.java:33)
	at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:266)
	at org.sireum.util.jvm.Exec.run(Exec.scala:100)
        ...
```

Steps to reproduce using 1.2.2 (note: this first downloads Zulu JDK, Sbt, and other dependencies):

```bash
git clone --recursive --depth=1 https://github.com/sireum/v3 sireum-v3
sed -i.bak 's/nuprocess=1.1.3/nuprocess=1.2.2/' sireum-v3/versions.properties
sireum-v3/bin/sbt-launch.sh "project logika-jvm" test
```

robby-phd:
Thanks.  I can confirm that 1.2.2 on Linux with Zulu JDK 10 crashes:

```bash
java.lang.NoClassDefFoundError: Could not initialize class com.zaxxer.nuprocess.internal.LibC
	at com.zaxxer.nuprocess.internal.BasePosixProcess.createPipes(BasePosixProcess.java:585)
	at com.zaxxer.nuprocess.linux.LinuxProcess.start(LinuxProcess.java:86)
	at com.zaxxer.nuprocess.linux.LinProcessFactory.createProcess(LinProcessFactory.java:40)
	at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:266)
	at org.sireum.util.jvm.Exec.run(Exec.scala:100)
```

1.1.3 on Linux with the Zulu JDK works; 1.2.2 on Linux with Oracle Java 8 also works.

brettwooldridge:
@robby-phd NuProcess v1.2.3 has been published with support for Azul Systems' Zulu JVM.

robby-phd:
Great, thanks!

PR #107, opened 2020-01-16
Don't special case Azul JVMs.

bturner:
On #87 (May 2018), there was a report of a linkage issue when using Azul Zulu 10. As part of fixing that issue, a check was added to use a new code path when the JVM vendor included "Azul".

A month later, #91 (June 2018) was created showing a similar linkage error when using OpenJDK 10. For that issue, the Azul-specific change from #87 was generalized to apply to any Java 10+ JVM, but the Azul vendor check was not removed. As a result, Azul Zulu 8 (for example) attempts to use the Java 10 code path and fails.

- Removed the Azul vendor check. Now, regardless of JVM vendor:
  - Java 10+ JVMs use `Java_java_lang_ProcessImpl_forkAndExec`
  - Earlier JVMs use `Java_java_lang_UNIXProcess_forkAndExec`

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
@@ -35,11 +35,7 @@ import static com.zaxxer.nuprocess.internal.Constants.JVM_MAJOR_VERSION;
  */
 public class LinuxProcess extends BasePosixProcess
 {
-   private static final boolean isAzul;
-
    static {
-      isAzul = System.getProperty("java.vm.vendor", "").contains("Azul");
-
       LibEpoll.sigignore(LibEpoll.SIGPIPE);
 
       // TODO: install signal handler for SIGCHLD, and call onExit() when received, call the default (JVM) hook if the PID is not ours
@@ -90,7 +86,7 @@ public class LinuxProcess extends BasePosixProcess
          createPipes();
          int[] child_fds = {stdinWidow, stdoutWidow, stderrWidow};
 
-         if (JVM_MAJOR_VERSION >= 10 || isAzul) {
+         if (JVM_MAJOR_VERSION >= 10) {
             pid = com.zaxxer.nuprocess.internal.LibJava10.Java_java_lang_ProcessImpl_forkAndExec(
                   JNIEnv.CURRENT,
                   this,


bturner:
I've tested this with:
* Amazon Corretto 8u242
* Amazon Corretto 11u6
* Azul 7.36 (7u252)
* Azul 8.12 (8u71)
* Azul 8.44 (8u242)
* Azul 11.37 (11u6)
* OpenJDK 8u232
* OpenJDK 11u5
* Oracle JDK 8u201
* Oracle JDK 10u2
* Oracle JDK 11u6

Without this change, Azul 7.36, 8.12 and 8.44 all fail. With it, every version tested above passes.

Note: Since #87 mentioned macOS and Azul, I've re-tested Azul 8.44 and 11.37 there just for a little extra validation; no issues. (I know `LinuxProcess` isn't used on macOS; I re-tested just to ensure I hadn't somehow regressed #87.)

bturner:
As part of preparing to move Atlassian Bitbucket Server over to using NuProcess to run `git` processes, rather than Java's `ProcessBuilder`, we introduced a check in our 6.9.0 release which uses NuProcess to run `java -version` and displays an alert in the UI if that fails, to help detect installations on JVMs, or platforms, where NuProcess won't work.

As a result of that check, we've found that RHEL 5.x can't be used with NuProcess because JNA requires glibc 2.7 or newer (it might be worth noting this in NuProcess's README, in addition to the kernel minimum version). That's unfortunate, but not entirely unexpected; RHEL 5 is very nearly end-of-life (November 2020).

However, we also had a support case from a customer running RHEL 7.6, which has a sufficiently new kernel (3.10) and glibc (2.17). Investigating their environment revealed they're using Zulu 8.41 (8u231), which brought us to this issue in NuProcess. We're tracking this issue as [BSERV-12135](https://jira.atlassian.com/browse/BSERV-12135).

benhumphreys:
In addition to the above testing with Azul's Zulu, I've tested with their Zing JVM too.

**Java 8**
```
java version "1.8.0-zing_19.12.100.0"
Zing Runtime Environment for Java Applications (build 1.8.0-zing_19.12.100.0-b1)
Zing 64-Bit Tiered VM (build 1.8.0-zing_19.12.100.0-b1-product-linux-X86_64, mixed mode)
```

With the above version, NuProcess master fails a `mvn verify` with error:
```
java.lang.NoClassDefFoundError: Could not initialize class com.zaxxer.nuprocess.internal.LibJava10
	at com.zaxxer.nuprocess.EnvironmentTest.runNuProcess(EnvironmentTest.java:86)
	at com.zaxxer.nuprocess.EnvironmentTest.defaultEnv(EnvironmentTest.java:62)
```
While including this change `mvn verify` passes.

**Java 11**
```
java version "11.0.5.0.101" 2020-01-07 LTS
Zing Runtime Environment for Java Applications 19.12.100.0+1 (product build 11.0.5.0.101+11-LTS)
Zing 64-Bit Tiered VM 19.12.100.0+1 (product build 11.0.5.0.101-zing_19.12.100.0-b1-product-linux-X86_64, mixed mode)
```
Both `master`, and `bturner:bturner-azul-java-8` pass.

brettwooldridge:
Awesome! Thanks for the deep investigation and fixes!

brettwooldridge:
@bturner NuProcess v1.2.6 has been published.

PR #109, opened 2020-03-29
fixed SYS___pthread_chdir value for FreeBSD when making call to syscall

jasonkafer:
Fix for FreeBSD LibC chdir syscal

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
@@ -27,14 +27,24 @@ import static com.zaxxer.nuprocess.internal.Constants.OS;
 @SuppressWarnings("WeakerAccess")
 public class LibC
 {
+    
+   public static int SYS___pthread_chdir = 348; 
+    
    static {
       if (OS == OperatingSystem.MAC) {
-         O_NONBLOCK = 0x0004; // MacOS X, Freebsd
+         O_NONBLOCK = 0x0004; // MacOS X, Freebsd 
       }
+      
       else {
          O_NONBLOCK = 2048; // Linux
       }
-
+      
+      // changing the value based on https://github.com/freebsd/freebsd/blob/master/sys/sys/syscall.h
+      String osname = System.getProperty("os.name").toLowerCase();
+      if ( osname.contains("freebsd") ) {
+          SYS___pthread_chdir = 12;
+      }
+       
       Native.register(NativeLibrary.getProcess());
    }
 
@@ -81,8 +91,7 @@ public class LibC
    // We can't use JNA direct mapping for syscall(), since it takes varargs.
    public interface SyscallLibrary extends Library
    {
-      public static final int SYS___pthread_chdir = 348;
-
+     
       int syscall(int syscall_number, Object... args);
    }
 
--- a/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
@@ -34,7 +34,6 @@ public class LibC
       if (OS == OperatingSystem.MAC) {
          O_NONBLOCK = 0x0004; // MacOS X, Freebsd
       }
-      
       else {
          O_NONBLOCK = 2048; // Linux
       }
--- a/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/LibC.java
@@ -32,7 +32,7 @@ public class LibC
     
    static {
       if (OS == OperatingSystem.MAC) {
-         O_NONBLOCK = 0x0004; // MacOS X, Freebsd 
+         O_NONBLOCK = 0x0004; // MacOS X, Freebsd
       }
       
       else {
--- a/src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
@@ -26,7 +26,6 @@ import com.zaxxer.nuprocess.NuProcessHandler;
 import com.zaxxer.nuprocess.internal.BasePosixProcess;
 import com.zaxxer.nuprocess.internal.IEventProcessor;
 import com.zaxxer.nuprocess.internal.LibC;
-import com.zaxxer.nuprocess.internal.LibC.SyscallLibrary;
 
 import java.nio.file.Path;
 import java.util.List;
--- a/src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
@@ -181,14 +181,14 @@ class OsxProcess extends BasePosixProcess
       Pointer oldCwd = new Pointer(peer);
       LibC.getcwd(oldCwd, cwdBufSize);
       String newCwd = cwd.toAbsolutePath().toString();
-      int rc = LibC.SYSCALL.syscall(SyscallLibrary.SYS___pthread_chdir, newCwd);
+      int rc = LibC.SYSCALL.syscall(LibC.SYS___pthread_chdir, newCwd);
       checkReturnCode(rc, "syscall(SYS__pthread_chdir) failed to set current directory");
 
       try {
          return LibC.posix_spawnp(restrict_pid, restrict_path, file_actions, restrict_attrp, argv, envp);
       }
       finally {
-         rc = LibC.SYSCALL.syscall(SyscallLibrary.SYS___pthread_chdir, oldCwd);
+         rc = LibC.SYSCALL.syscall(LibC.SYS___pthread_chdir, oldCwd);
          Native.free(Pointer.nativeValue(oldCwd));
          checkReturnCode(rc, "syscall(SYS__pthread_chdir) failed to restore current directory");
       }


src/main/java/com/zaxxer/nuprocess/internal/LibC.java
```diff
@@ -27,14 +27,24 @@
 @SuppressWarnings("WeakerAccess")
 public class LibC
 {
+    
+   public static int SYS___pthread_chdir = 348; 
+    
    static {
       if (OS == OperatingSystem.MAC) {
-         O_NONBLOCK = 0x0004; // MacOS X, Freebsd
+         O_NONBLOCK = 0x0004; // MacOS X, Freebsd
```
brettwooldridge: Thanks for the pull request, but please remove unrelated whitespace changes.

src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
```diff
@@ -181,14 +181,14 @@ private int spawnWithCwd(final IntByReference restrict_pid,
...
-      int rc = LibC.SYSCALL.syscall(SyscallLibrary.SYS___pthread_chdir, newCwd);
+      int rc = LibC.SYSCALL.syscall(LibC.SYS___pthread_chdir, newCwd);
       checkReturnCode(rc, "syscall(SYS__pthread_chdir) failed to set current directory");
 
       try {
          return LibC.posix_spawnp(restrict_pid, restrict_path, file_actions, restrict_attrp, argv, envp);
       }
       finally {
-         rc = LibC.SYSCALL.syscall(SyscallLibrary.SYS___pthread_chdir, oldCwd);
+         rc = LibC.SYSCALL.syscall(LibC.SYS___pthread_chdir, oldCwd);
```
brettwooldridge: Have you tested the above two changes on MacOS X to ensure compatibility?

src/main/java/com/zaxxer/nuprocess/internal/LibC.java
jasonkafer: Whitespace has been removed, sorry about that.

src/main/java/com/zaxxer/nuprocess/osx/OsxProcess.java
jasonkafer: The test case RunTest.changeCwd() passes on macOS 10.15.4, it appears that the syscall.h for OSX has the value of 348 specified (https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.15.sdk/usr/include/sys/syscall.h) whereas the FreeBSD syscall.h does not.  Also worth noting that the FreeBSD value in syscall.h is SYS_chdir, there is not a SYS___pthread_chdir in the header like there is in the OSX version.  I am not sure of the impact of using the traditional SYS_chdir vs SYS___pthread_chdir on FreeBSD, but I can confirm the change does work on FreeBSD 11.3.

brettwooldridge: Given clearer understanding of the change, I actually expect issues under load.  However, there is no multi-threaded load-test version of the changeCwd() test.  One needs to be written.

Here is how the current (pthread) code works on OS X:
```
Thread 1: pthread_chdir /pathA
Thread 1: spawn process ... process inherits cwd of /pathA
```
Now multi-threaded interleaved:
```
Thread 1: pthread_chdir /pathA
Thread 2: pthread_chdir /pathB
Thread 1: spawn process ... process inherits cwd of /pathA
Thread 2: spawn process ... process inherits cwd of /pathB
```
Because SYS_chdir is global rather than thread specific, here what I expect to happen with multiple interleaved threads:
```
Thread 1: SYS_chdir /pathA
Thread 2: SYS_chdir /pathB
Thread 1: spawn process ... process inherits cwd of /pathB
Thread 2: spawn process ... process inherits cwd of /pathB
```
The process spawned by thread 1 will have the incorrect cwd.

I would need to see a test that convinces me that is not the case before I can accept the change.

jasonkafer: Brett -

I'll work on a test case to implement the above, I did quite a bit of digging through the FreeBSD source to find a pthread equivalent chdir call, and couldn't find anything, more research to be done, but I wonder if somehow freebsd is handling this differently.  We've replaced the library in bitbucket with the version that implements this change, and it is a quite active server and we have thus far not experienced any issues. 

But as I said, I'll implement the test case and run the unit tests on FreeBSD and see what the results are. 

Jason

brettwooldridge: Alternatively, you can add a conditional path in the code, one for FreeBSD and one for OSX, and roll the dice on FreeBSD.

jasonkafer: That is more or less what I am doing; i'm only changing the value of SYS___pthread_chdir to 12 in LibC if the os is Freebsd, otherwise it remains 348 (including on Mac), so in a sense, I'm rolling the dice now 😊 

I feel like there is more to this puzzle with freebsd and pthread, and just haven't had the time to dig further into it... I agree with your test case and think that is a good spot to start. If the multithreaded load test proves successful on FreeBSD it would indicate its doing something under the hood with pthread and somehow mapping the SYS_chdir to the executing thread as oppose to the process... If it fails, then I'm missing something and need to dig deeper. Unfortunately the standard simple google queries provide no details on this, so its a dig through the FreeBSD source kind of issue, my favorite haha

src/main/java/com/zaxxer/nuprocess/internal/LibC.java
```diff
@@ -81,8 +90,7 @@ public static native int posix_spawnp(IntByReference restrict_pid, String restri
    // We can't use JNA direct mapping for syscall(), since it takes varargs.
    public interface SyscallLibrary extends Library
    {
-      public static final int SYS___pthread_chdir = 348;
-
+
```
bturner, line 93: If we don't put the `chdir` constant back here, should this line be removed?

src/main/java/com/zaxxer/nuprocess/internal/LibC.java
```diff
@@ -27,14 +27,23 @@
 @SuppressWarnings("WeakerAccess")
 public class LibC
 {
+    
+   public static int SYS___pthread_chdir = 348;
```
bturner, line 31: It doesn't seem like this should be mutable. This is clearly intended to be a constant.

Rather than initializing this here and changing it in the `static` block, which requires this to be mutable, leave it uninitialized here and update the `static` block to set the "right" value:
```java
public static final int SYS___pthread_chdir;

static {
   // O_NONBLOCK initialization as now

   SYS___pthread_chdir = System.getProperty("os.name").toLowerCase().contains("freebsd") ? 12 : 348;

   // Register Native
}
```

If this constant is going to move to `LibC`, should it be further down, with the others (like `O_NONBLOCK`)? Alternatively, perhaps the constant should just stay where it was, paired with a new `static` block inside `SyscallLibrary` to initialize it?

skissane:
I just wanted to make the point that `SYS__pthread_chdir` and `SYS__chdir` do really different things – the first sets the per-thread working directory, the second sets the per-process working directory. FreeBSD *does not support per-thread working directory*. Very few platforms do. In fact, the only platforms I know of that do are Linux and XNU (but Linux's implementation is very different from, and incompatible with, XNU's). 

Setting `SYS__pthread_chdir = 12` on FreeBSD is going to really confuse people because the constant name implies "per-thread working directory" but you are setting it to a value which means "per-process working directory". It is also going to introduce thread safety issues when replacing a thread-safe API with a non-thread-safe one.

It is *possible* that FreeBSD developers might some day add support for `pthread_chdir_np`/`pthread_fchdir_np`. On the one hand, they'll likely argue that it is the wrong approach, and `openat`/etc are the right approach. Especially given FreeBSD supports that approach in scenarios many other platforms don't (e.g. `bindat`/`connectat` for Unix domain sockets). OTOH, making it easier to port software from macOS/XNU might convince them.

But, in the meantime, I think the whole approach here is wrong – since NuProcesses' macOS process startup code is based on `SYS__pthread_chdir`, it is not a good base for a FreeBSD implementation.

A better approach could be something like this:
- Refactor `OsxProcess` into an abstract base class `BsdProcess`, with two concrete subclasses – `OsxProcess` and `FreeBsdProcess`
- `OsxProcessFactory` will decide at runtime whether to create an `OsxProcess` or a `FreeBsdProcess` depending on what the OS is (possibly should rename it to `BsdProcessFactory`, but that might break backward compatibility – one option might be to do the rename, but keep `OsxProcessFactory` as a dummy subclass)
- Make the `spawnWithCwd` method of `BsdProcess` be `protected abstract`, so the concrete subclasses have to implement their own `spawnWithCwd`
- For the `FreeBsdProcess` implementation of `spawnWithCwd`, use a completely different strategy. For example, you could spawn an external helper binary written in C, which does a chdir() to its first argument, then execs the remaining arguments.
- An alternative to relying on an external helper binary: `LinuxProcess` does this (using the `jspawnhelper` binary shipped with the JDK) – I'm not sure if that code works on FreeBSD, but if it does you could base `FreeBsdProcess` more off `LinuxProcess` than `OsxProcess`.
- You could even possibly use `/bin/sh` as in this: https://unix.stackexchange.com/a/480779
- As that link notes, there are heaps of third party utilities which can do this (cd and exec), but while some of them may be in FreeBSD ports, you can't rely on them being there on a random FreeBSD system. Although, that might be some argument for a system property/etc to let you specify an arbitrary utility.

bturner:
Thanks for the detailed feedback, @skissane! I'll readily admit, I lack the depth of knowledge you have around FreeBSD and how some of these internals work there. Looking purely at end behavior, though, trying to use the per-process work directory to control working directories for processes we spawn, and can actively be spawning on multiple threads concurrently, seems like an obvious no-no. (I no longer work on Bitbucket Server, but thinking about this in terms of Bitbucket Server's usage of the library it'd be pretty disastrous to, for example, have 2 threads that are trying to operate on different repositories end up racily changing the process-wide working directory and one thread's update stomps the other, resulting in the forked `git` processes both running on the same repository. The support case to try and find that issue would be brutal.)

Based on that potential issue, and reinforced given that I asked for a variety of rework on this pull request over 3 years ago and it hasn't been actioned, I think the right approach is to decline this change and consider other options separately. One thing I'll add to that, though, is that I don't a) use FreeBSD or b) have any sort of FreeBSD setup, so I personally am unlikely to make any inroads on a solution to this issue.

Issue #112, opened 2020-05-19
wantWrite() race


bturner:
Consider a situation where NuProcess is managing a process to/from which input and output are being pumped to a remote connection (for example an HTTP or SSH request) which is using non-blocking I/O.

In such a setup, `wantWrite()` would be called when the remote client sends some data that should be forwarded `stdin` (perhaps triggered by something like Servlet 3.1's `ReadListener.onDataReady`), and then `onStdinReady` would be called when NuProcess is ready to read and forward those bytes. `onStdinReady` may get invoked some number of times, if there's a lot of data available, but eventually that data runs out. At that point, `onStdinReady` will return `false` (because there's no more data ready for `stdin` at that time), and the process will wait for another `wantWrite()` when more data is received.

Unfortunately, the transition between `onStdinReady` and `wantWrite` is racy, when `wantWrite` is being called from some thread other than the NuProcess pump thread (and it almost has to be, given the pump thread is blocked in `epoll_wait` or similar). Specifically, [this assignment](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java#L486-L487) is problematic:
```java
         boolean wantMore = processHandler.onStdinReady(inBuffer);
         userWantsWrite.set(wantMore);
```

If the `wantWrite()` call occurs "between" `onStdinReady` returning and the call to `userWantsWrite.set(wantMore)`, `wantWrite()` sets `userWantsWrite` to `true` and then it's immediately set back to `false` by `wantMore`. This leads to a hang, because the HTTP/SSH request has signaled that it wanted to write and is waiting for data to be moved, but `onStdinReady` will never be called because the process has accidentally cleared the `userWantsWrite` flag.

bturner:
In Servlet 3.1, `ReadListener.onDataReady` is called exactly once, when a the web container's internal buffer goes from no data available to some. At that point it won't be called again until `ServletInputStream.isReady` has been called and returned `false`, per its contract:
```java
    /**
     * When an instance of the <code>ReadListener</code> is registered with a {@link ServletInputStream},
     * this method will be invoked by the container the first time when it is possible
     * to read data. Subsequently the container will invoke this method if and only
     * if {@link javax.servlet.ServletInputStream#isReady()} method
     * has been called and has returned <code>false</code>.
     *
     * @throws IOException if an I/O related error has occurred during processing
     */
    public void onDataAvailable() throws IOException;
```

[...]

PR #113, opened 2020-05-26
Clear userWantsWrite before calling onStdinReady. (#112)

bturner:
Prior to this change, if `wantWrite()` was called from another thread while NuProcess was calling `onStdinReady(ByteBuffer)`, that request would be "lost" if `onStdinReady` returned `false`. Depending on where input is coming from (for example a Servlet 3.1 `ReadListener`), that could result in the process getting "stuck" indefinitely.

- Updated `BasePosixProcess.writeStdin(int, int)` to clear the `userWantsWrite` flag, if it's set, prior to calling `onStdinReady`, and then only re-set `userWantsWrite` if `onStdinReady` returns `true`
- Updated `WindowsProcess.writeStdin(int)` to use the same technique
  - Also fixed an unrelated bug where `checkHandleValidity` was comparing the provided `HANDLE` to the invalid handle's pointer, rather than comparing pointers from both
- Added a new unit test that verifies the `wantWrite()` request isn't "lost" if it happens during an `onStdinReady` callback
  - This test fails on Linux, macOS and Windows without the changes to `writeStdin` in `BasePosixProcess` and `WindowsProcess`

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -477,20 +477,27 @@ public abstract class BasePosixProcess implements NuProcess
          }
       }
 
-      if (!userWantsWrite.get()) {
+      // Check whether the user has requested a write window. If so, this will clear the request and set
+      // userWantsWrite to a known-false state. Clearing the flag here ensures, if wantWrite() is called
+      // while we're in onStdinReady or, racily, after onStdinReady returns but before the result can be
+      // used to update userWantsWrite, we don't "lose" the write request
+      if (!userWantsWrite.compareAndSet(true, false)) {
          return false;
       }
 
       try {
          inBuffer.clear();
-         boolean wantMore = processHandler.onStdinReady(inBuffer);
-         userWantsWrite.set(wantMore);
+         if (processHandler.onStdinReady(inBuffer)) {
+            // If onStdinReady returns true, re-set userWantsWrite. If it returns false, we already set
+            // userWantsWrite to false before we made the callback so there's nothing to do
+            userWantsWrite.set(true);
+         }
          if (inBuffer.hasRemaining() && availability > 0) {
             // Recurse
             return writeStdin(availability, fd);
-         } else {
-            return true;
          }
+
+         return true;
       }
       catch (Exception e) {
          Logger.getLogger(getClass().getName()).log(Level.SEVERE, "Exception thrown handling writes to stdin " + processHandler, e);
--- a/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
@@ -33,6 +33,8 @@ import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import static com.zaxxer.nuprocess.internal.Constants.NUMBER_OF_THREADS;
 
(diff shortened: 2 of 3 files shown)


src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -437,23 +439,27 @@ boolean writeStdin(int transferred)
          }
       }
 
-      if (userWantsWrite.compareAndSet(true, false)) {
```
bturner, line 440: `WindowsProcess` was already using the `compareAndSet` to clear the old request, but paired with a blind `userWantsWrite.set` of the `onStdinReady` result, which means `WindowsProcess` still exhibited the race.

I've flipped the handling here (`!userWantsWrite.compareAndSet`) to make this method and `BasePosixProcess.writeStdin(int, int)` flow more similarly. I'm happy to revert that if you prefer.

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -712,7 +718,7 @@ private String getEnvironmentBlock(Map<String, String> env)
 
    private void checkHandleValidity(HANDLE handle)
    {
-      if (NuWinNT.INVALID_HANDLE_VALUE.getPointer().equals(handle)) {
+      if (NuWinNT.INVALID_HANDLE_VALUE.getPointer().equals(handle.getPointer())) {
```
bturner, line 721: This was an unrelated issue I noticed while in the code. I don't see how the previous code could ever have correctly detected an invalid handle

src/test/java/com/zaxxer/nuprocess/CatTest.java
```diff
@@ -301,6 +301,69 @@ public void softCloseStdinAfterWrite() throws Exception
...
+      nuProcess.wantWrite();
+      //Wait until onStdinReady is called
+      onStdinReadyCalled.await(5L, TimeUnit.SECONDS);
+      //Call wantWrite() again
+      nuProcess.wantWrite();
+      //Let the first onStdinReady callback complete
+      wantWriteCalled.countDown();
+
+      try {
+         Assert.assertNotEquals(command + " did not complete",
```
bturner, line 354: Without the `writeStdin` changes, this `waitFor` times out and the process is ultimately killed (so the test doesn't hang indefinitely on a breakage). The `callCount` verification is just an added safety net to try and ensure the test is passing for the right reason.

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -477,20 +477,27 @@ else if (byteBuffer != null) {
...
       }
 
       try {
          inBuffer.clear();
-         boolean wantMore = processHandler.onStdinReady(inBuffer);
-         userWantsWrite.set(wantMore);
+         if (processHandler.onStdinReady(inBuffer)) {
+            // If onStdinReady returns true, re-set userWantsWrite. If it returns false, we already set
+            // userWantsWrite to false before we made the callback so there's nothing to do
+            userWantsWrite.set(true);
```
bturner, line 493: With this approach, if a user requests a write window, we clear that request and then call `onStdinReady`. If another call to `wantWrite` happens during `onStdinReady`, _regardless of what `onStdinReady` returns_, you're guaranteed another callback.

It's possible that the in-flight `onStdinReady` will actually have handled the data that the concurrent `wantWrite` wanted to send, depending on how the program is structured, which means this could result in an "extra" callback. That's unfortunate, but writing code inside a handler to prevent it, or handle it, is not that complex. The counter case, if this change isn't made, is much worse: If `onStdinReady` returns `false`, and a `wantWrite()` request happens concurrently, no `onStdinReady` callback actually happens. More importantly, it's not possible inside a handler to "fix" this race condition.

When paired with a `ReadListener` from a Java `ServletInputStream`, this can hang the process. When `ReadListener.onDataAvailable` is called, it will _never_ be called again until `isReady()` has been called and returned `false`. That means if its `wantWrite()` is dropped because it's timed just wrong with an `onStdinReady` callback (rare, certainly, but not really all that difficult to achieve in practice), `onStdinReady` is never called an `isReady` may never happen, so no subsequent `wantWrite()` will retry it.

bturner:
Thanks for the quick merge and release, @brettwooldridge. I appreciate all your support.

brettwooldridge:
@bturner No, thank *you* for the contributions.

PR #118, opened 2020-09-25
Reduce pipe buffer size on Windows (fixes #117).

### LINKED ISSUE #117: Stdout and stderr handling can be extremely slow on Windows
When used to pump processes that produce lots of output in small chunks (which happens a lot with `git` commands--especially `git http-backend` and `git upload-pack`, which are used to serve clones and fetches), the existing I/O completion handler code in `ProcessCompletions` runs afoul of how IOCP decides when to signal ready reads. When some data is available to read, IOCP waits (a very short wait) to see if more data arrives before triggering. For short-lived processes, or processes that produce their output in big chunks, that works fine. But if the process produces output in many, many tiny pieces, that extra delay amounts to a huge performance hit.

This came up for Bitbucket Server in [BSERV-12599](https://jira.atlassian.com/browse/BSERV-12599), where it was reported that, since we switched over to NuProcess to run `git http-backend` and `git upload-pack`, hosting operations on Windows are an order of magnitude slower than they were using `ProcessBuilder` and blocking I/O.

To give a sense of scale, cloning a 500MB repository (Bitbucket Server's own source) via Bitbucket Server using `ProcessBuilder` looks like this:
```
C:\Temp>git clone --no-checkout http://localhost:7990/bitbucket/scm/at/bitbucket-server.git
Cloning into 'bitbucket-server'...
remote: Enumerating objects: 2304995, done.
remote: Counting objects: 100% (2304995/2304995), done.
remote: Compressing objects: 100% (779774/779774), done.
remote: Total 2304995 (delta 1058323), reused 2304995 (delta 1058323), pack-reused 0
Receiving objects: 100% (2304995/2304995), 480.51 MiB | 33.96 MiB/s, done.
Resolving deltas: 100% (1058323/1058323), done.
```

480MB at 34MB/s, with the entire operation taking about 25 seconds (the 34MB/s transfer is only part of the overall time).

Switching over to NuProcess completely tanks performance:
```
C:\Temp>git clone --no-checkout http://localhost:7990/bitbucket/scm/at/bitbucket-server.git
Cloning into 'bitbucket-server'...
remote: Enumerating objects: 2304995, done.
remote: Counting objects: 100% (2304995/2304995), done.
remote: Compressing objects: 100% (779774/779774), done.
remote: Total 2304995 (delta 1058323), reused 2304995 (delta 1058323), pack-reused 0
Receiving objects: 100% (2304995/2304995), 480.51 MiB | 2.08 MiB/s, done.
Resolving deltas: 100% (1058323/1058323), done.
```

We've dropped from 34MB/s to 2MB/s, and the overall operation now takes over 3 minutes. For larger repositories the difference is even more painful, taking clones that previously ran in 30-60 seconds and blowing them out to 10-15 minutes. That results in stacking load on Bitbucket Server that eventually causes rejected requests due to excessive queuing.

I stripped out all Bitbucket Server's code and wrote a test in NuProcess that runs `git http-backend` directly, with the right `stdin` and environment to produce the same effective operation. (Unfortunately this test isn't really _shareable_ because it relies on some canned `stdin` I captured, as well as access to a specific Git repository.) With that test, I'm able to reproduce the performance issue without any Bitbucket Server code at all. (It's worth noting that the test executed on Linux or macOS performs fine, with NuProcess speeds essentially identical to `ProcessBuilder`.)

In trying to track down the issue, I looked through the JDK's source and found [they use `4096 + 24` byte buffers](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/windows/native/java/lang/ProcessImpl_md.c#L36-L41) for their pipes. Changing [`WindowsProcess.BUFFER_SIZE`](https://github.com/brettwooldridge/NuProcess/blob/nuprocess-2.0.1/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java#L48) from 64K to `4096 + 24` fixes the issue and produces identical throughput with NuProcess compared to `ProcessBuilder`.

A colleague helping me search for this found some other cases where IOCP's Nagle-like approach has caused problems:
* https://news.ycombinator.com/item?id=19520953
* https://github.com/python-trio/trio/issues/52
  * This one links off to several issues Chrome developers have run into related to IOCP performance

bturner:
Using a large buffer size for the pipe connecting the child process can result in IOCP imposing (very short) delays before notifying the port when a read completes, waiting in case additional data arrives. For a small number of reads the short delays are no problem, but for processes that produce hundreds of megabytes of output (especially if they do so in small bursts, like `git http-backend` and `git upload-pack`) all those short delays accumulate into a massive performance hit, reducing throughput compared to `ProcessBuilder` by an order of magnitude (30MB/s+ down to 2MB/s).

Using a smaller pipe buffer does impose some constraint on how much data can be moved by a single `ReadFile` or `WriteFile` call but it also prevents delays (because there's no point in waiting for more input if the buffer is already full), resulting in overall improved--and, more importantly, more consistent--performance.

- Reduced `WindowsProcess.BUFFER_SIZE` from 64K to 4096+24, which matches the buffer size the JDK's `ProcessImpl` uses on Windows
- Replaced `System.err` with a `Logger` when writing errors
  - This matches how `LinuxProcess` and `OsxProcess` are written
- Optimized `HANDLE.fromNative` to check the pointer directly, to avoid creating `HANDLE` instances for invalid handles
  - This also avoids the base implementation's use of reflection to instantiate new `HANDLE` instances
- Updated `PipeBundle` to set auto-sync to false on `OVERLAPPED` instances so JNA won't waste time marshaling to/from native code around every call to `ReadFile` or `WriteFile`

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/windows/NuWinNT.java
+++ b/src/main/java/com/zaxxer/nuprocess/windows/NuWinNT.java
@@ -58,10 +58,12 @@ public interface NuWinNT
 
    int STARTF_USESTDHANDLES = 0x100;
 
-   HANDLE INVALID_HANDLE_VALUE = new HANDLE(Pointer.createConstant(Native.POINTER_SIZE == 8 ? -1 : 0xFFFFFFFFL));
+   HANDLE INVALID_HANDLE_VALUE = new HANDLE(HANDLE.INVALID);
 
    class HANDLE extends PointerType
    {
+      static final Pointer INVALID = Pointer.createConstant(Native.POINTER_SIZE == 8 ? -1 : 0xFFFFFFFFL);
+
       public HANDLE()
       {
       }
@@ -74,15 +76,15 @@ public interface NuWinNT
       @Override
       public Object fromNative(Object nativeValue, FromNativeContext context)
       {
-         Object o = super.fromNative(nativeValue, context);
-         if (INVALID_HANDLE_VALUE.equals(o)) {
+         Pointer ptr = (Pointer) nativeValue;
+         if (INVALID.equals(ptr)) {
             return INVALID_HANDLE_VALUE;
          }
-         return o;
+         return new HANDLE(ptr);
       }
    }
 
-   static class WORD extends IntegerType
+   class WORD extends IntegerType
    {
       public static final int SIZE = 2;
 
@@ -97,7 +99,7 @@ public interface NuWinNT
       }
    }
 
-   static class DWORD extends IntegerType
+   class DWORD extends IntegerType
    {
       public static final int SIZE = 4;
 
@@ -112,7 +114,7 @@ public interface NuWinNT
       }
    }
 
-   static class ULONG_PTR extends IntegerType
+   class ULONG_PTR extends IntegerType
    {
       public ULONG_PTR()
       {
@@ -123,14 +125,9 @@ public interface NuWinNT
       {
          super(Native.POINTER_SIZE, value, true);
       }
-
-      public Pointer toPointer()
-      {
-         return Pointer.createConstant(longValue());
-      }
    }
 
-   static class ULONG_PTRByReference extends ByReference
+   class ULONG_PTRByReference extends ByReference
    {
       public ULONG_PTRByReference()
       {
@@ -166,10 +163,9 @@ public interface NuWinNT
       public boolean bInheritHandle;
 
       @Override
-      @SuppressWarnings("rawtypes")
-      protected List getFieldOrder()
+      protected List<String> getFieldOrder()
       {
-         return Arrays.asList(new String[] { "dwLength", "lpSecurityDescriptor", "bInheritHandle" });
+         return Arrays.asList("dwLength", "lpSecurityDescriptor", "bInheritHandle");
       }
    }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -45,8 +45,10 @@
 {
    private static final boolean IS_SOFTEXIT_DETECTION;
 
-   private static final int BUFFER_SIZE = 65536;
+   // See https://github.com/JetBrains/jdk8u_jdk/blob/master/src/windows/native/java/lang/ProcessImpl_md.c#L36-L41
+   private static final int BUFFER_SIZE = 4096 + 24;
```
bturner, line 49: If you'd prefer, I could also change this to be something like:
```java
private static final int BUFFER_SIZE = Integer.parseInt(System.getProperty("com.zaxxer.nuprocess.pipebuffer", "65536"));
```

That would preserve the existing large buffer, but give me the option of configuring a smaller value in Bitbucket Server using a system property. In practice, though, I suspect very few applications will notice the difference in this change to the default. I've run testing that forks thousands of processes and the test duration is identical overall before and after reducing the buffer size.

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -777,7 +780,12 @@ public int compare(Map.Entry<String, String> e1, Map.Entry<String, String> e2)
       {
          this.pipeHandle = pipeHandle;
          this.ioCompletionKey = ioCompletionKey;
-         this.overlapped = new OVERLAPPED();
+
+         // The OVERLAPPED structure is required to make non-blocking ReadFile and WriteFile calls,
+         // but its state is never read from nor written to in Java, so we disable auto-sync so JNA
+         // doesn't waste time marshaling contents to/from native memory
+         overlapped = new OVERLAPPED();
+         overlapped.setAutoSynch(false);
```
bturner, line 788: In my testing, which involved a process that took ~122,000 `ReadFile` calls to complete, disabling auto-sync on `OVERLAPPED` cut ~20 seconds off the overall execution time.

src/main/java/com/zaxxer/nuprocess/windows/NuWinNT.java
```diff
@@ -74,15 +76,15 @@ public HANDLE(Pointer p)
       @Override
       public Object fromNative(Object nativeValue, FromNativeContext context)
       {
-         Object o = super.fromNative(nativeValue, context);
```
bturner, line 77: `super.fromNative` casts `nativeValue` to a `Pointer` and then constructs a new instance using reflection to call the nullary constructor, and then sets the pointer directly on the `pointer` field and returns it.

By doing the cast myself and comparing the pointer directly, I skip creating a full `HANDLE` just to throw it away if it's invalid. This also eliminates the `class.getDeclaredConstructor().newInstance()` reflection and just calls the `HANDLE(Pointer)` constructor directly.

That said, it's probably not super important for performance. `HANDLE` instances _do_ get marshaled from native a lot, though, so it may have some minimal impact.

sschuberth:
Any chance to get this merged @brettwooldridge? The unconditional `e.printStackTrace()` is highly disturbing.

PR #120, opened 2020-09-26
Ensure LinuxProcess closes pipes if starting a process fails (fixes #119)

### LINKED ISSUE #119: Descriptors/handles are leaked when processes fail to start
The exact codepaths vary, but for every supported platform there are cases where a failure to start a process leaks descriptors or handles.

For example, on Linux, if [`forkAndExec`](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java#L164) throws an exception, `closePipes()` is never called and the 3 `*Widow` descriptors are leaked. You can confirm this with NuProcess's own test suite:
* Put a breakpoint in [`RunTest.noExecutableFound()`](https://github.com/brettwooldridge/NuProcess/blob/master/src/test/java/com/zaxxer/nuprocess/RunTest.java#L219) after the call to `pb.start()`
* When the breakpoint is hit, run `lsof` on the `java` process

At the end of the `lsof` output will be something like the following (the IDs will likely vary, of course)
```
java    3858 bturner   24r     FIFO   0,12       0t0  119418 pipe
java    3858 bturner   27w     FIFO   0,12       0t0  119419 pipe
java    3858 bturner   29w     FIFO   0,12       0t0  119420 pipe
```

If you use the debugger to manually call `process.closePipes()` and then run `lsof` again, you'll see that the 3 pipes are closed.

That specific codepath is handled correctly on macOS and all the pipes are closed, but if there are any errors in `createPosixPipes()` after it calls `createPipes()`, since `createPosixPipes()` is called outside the `try`/`catch` in `run` and `start`, the pipes are leaked. (This case is much less likely that the Linux one, to be clear; I'm just noting its an edge case that exists.)

`WindowsProcess` has similar issues with handles it creates in `createPipes()`.

bturner:
- Fixed the `forkAndExec` native signatures to include `throws IOException` to match `ProcessImpl`'s version
  - This makes it more obvious that the calls in `LinuxProcess.prepareProcess` can throw on errors
- Moved calls to `LinuxProcess.closePipes()` from `run()` and `start()` into `prepareProcess` in a `try`/`finally` block around calls to the JVM's `forkAndExec` function
  - This way, if `forkAndExec` throws (which it does, on failure, as the signature change highlights) the `*Widow` descriptors are all closed
  - The non-`*Widow` ends of the pipes are closed in `onExit`, so they are already handled after failures
- Removed `BasePosixProcess.checkLaunch`
  - `OsxProcess` was the only place calling it, but `OsxProcess` didn't override it (so the call did nothing)
  - `LinuxProcess` override it, but since it was switched to using the JVM's `forkAndExec` function it no longer _calls_ it, so the override was just dead code
- Simplified `BasePosixProcess.initFailureCleanup`
- Updated some `BasePosixProcess` error paths to log their exceptions

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -89,11 +89,11 @@ public abstract class BasePosixProcess implements NuProcess
    private ConcurrentLinkedQueue<ByteBuffer> pendingWrites;
 
    static {
-      IS_SOFTEXIT_DETECTION = Boolean.valueOf(System.getProperty("com.zaxxer.nuprocess.softExitDetection", "true"));
+      IS_SOFTEXIT_DETECTION = Boolean.parseBoolean(System.getProperty("com.zaxxer.nuprocess.softExitDetection", "true"));
 
       processors = new IEventProcessor<?>[NUMBER_OF_THREADS];
 
-      if (Boolean.valueOf(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
+      if (Boolean.parseBoolean(System.getProperty("com.zaxxer.nuprocess.enableShutdownHook", "true"))) {
          Runtime.getRuntime().addShutdownHook(new Thread(new Runnable() {
             @Override
             public void run()
@@ -137,18 +137,6 @@ public abstract class BasePosixProcess implements NuProcess
     * @since 1.3
     */
    public abstract void run(List<String> command, String[] environment, Path cwd);
-
-   /**
-    * Check the launched process and return {@code true} if launch was successful,
-    * or {@code false} if there was an error in launch.
-    *
-    * @return {@code true} on success, {@code false} on failure
-    */
-   protected boolean checkLaunch()
-   {
-      // Can be overridden by subclasses for post-launch checks
-      return true;
-   }
 
    /** {@inheritDoc} */
    @Override
@@ -321,6 +309,7 @@ public abstract class BasePosixProcess implements NuProcess
       }
       catch (Exception e) {
          // Don't let an exception thrown from the user's handler interrupt us
+         LOGGER.log(Level.WARNING, "Exception thrown from handler", e);
       }
       finally {
          exitPending.countDown();
@@ -500,7 +489,7 @@ public abstract class BasePosixProcess implements NuProcess
          return true;
       }
       catch (Exception e) {
-         Logger.getLogger(getClass().getName()).log(Level.SEVERE, "Exception thrown handling writes to stdin " + processHandler, e);
+         LOGGER.log(Level.SEVERE, "Exception thrown handling writes to stdin " + processHandler, e);
 
          // Don't let an exception thrown from the user's handler interrupt us
          return false;
@@ -576,6 +565,7 @@ public abstract class BasePosixProcess implements NuProcess
       }
       catch (Exception e) {
          // Don't let an exception thrown from the user's handler interrupt us
+         LOGGER.log(Level.WARNING, "Exception thrown from handler", e);
       }
    }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -138,18 +138,6 @@ protected BasePosixProcess(NuProcessHandler processListener)
     */
    public abstract void run(List<String> command, String[] environment, Path cwd);
 
-   /**
-    * Check the launched process and return {@code true} if launch was successful,
-    * or {@code false} if there was an error in launch.
-    *
-    * @return {@code true} on success, {@code false} on failure
-    */
-   protected boolean checkLaunch()
```
bturner, line 147: Only `OsxProcess` was ever calling this, so it doesn't need to be in the base class

src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
```diff
@@ -627,25 +616,20 @@ protected void callStart()
    protected void initFailureCleanup(int[] in, int[] out, int[] err)
    {
       Set<Integer> unique = new HashSet<>();
-      if (in != null) {
```
bturner, line 630: The `in`, `out`, and `err` arrays can never be `null`; `createPipes` initializes all 3 before it makes the first native call.

src/main/java/com/zaxxer/nuprocess/internal/LibJava10.java
```diff
@@ -71,5 +72,5 @@ public static native int Java_java_lang_ProcessImpl_forkAndExec(
            Object dir,
            Object fds,
            byte redirectErrorStream
-   );
+   ) throws IOException;
```
bturner, line 75: This matches [Java's own declaration](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/solaris/classes/java/lang/UNIXProcess.java#L208-L215), which helps make it more obvious when `LinuxProcess` calls this method that it can throw.

src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
```diff
@@ -64,13 +63,10 @@ public NuProcess start(List<String> command, String[] environment, Path cwd) {
 
       try {
          prepareProcess(command, environment, cwd);
-
          if (pid == -1) {
```
bturner, line 66: I was tempted to remove this because it can never actually happen, as far as I can tell. If the JVM's `forkAndExec` fails to start the process, it always throws an exception--it doesn't return `-1` despite how the native code looks. But it doesn't hurt anything to have it here, so I left it.

src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
```diff
@@ -141,37 +134,44 @@ private void prepareProcess(List<String> command, String[] environment, Path cwd
       // See https://github.com/JetBrains/jdk8u_jdk/blob/master/src/solaris/classes/java/lang/ProcessImpl.java#L86
       byte[] envBlock = toEnvironmentBlock(environment);
 
-      // See https://github.com/JetBrains/jdk8u_jdk/blob/master/src/solaris/classes/java/lang/ProcessImpl.java#L96
```
bturner, line 144: I removed this comment because it's unrelated to the code it's attached to. The Java code now creates its own pipes, rather than using `{-1, -1, -1}` like it did when this comment was added.

I don't 100% understand why we can't take advantage of the pipe creation code in `forkAndExec`, but having tested going back to `{-1, -1, -1}` it seems like we can't. Most of the time it works, but randomly it seems to hang. Really bizarre. I can see that you [started with `{-1, -1, -1}`](https://github.com/brettwooldridge/NuProcess/commit/e838dbbba7a43be2458de4fc571cef62404ae0a6#diff-515a1963c8aa9b6ec932c230bed92969R90) but then [switched away to fix a "race condition seen once"](https://github.com/brettwooldridge/NuProcess/commit/dc9491e20e5b8db9170ac6e968bf9a1ebc48e78d#diff-515a1963c8aa9b6ec932c230bed92969R87), so it seems like you've already had that experience yourself. (I only tested switching back out of curiosity, just to be clear.)

src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
```diff
@@ -183,41 +183,6 @@ private void closePipes()
       LibC.close(stderrWidow);
    }
 
-   @Override
-   protected boolean checkLaunch()
```
bturner, line 187: This didn't show as unused because it was overriding a base class method, but it was never actually called

PR #123, opened 2021-05-12
Support Apple M1

changvvb:
Upgrade jna to support Apple M1. Ref https://github.com/java-native-access/jna/issues/1313


brettwooldridge:
@changvvb Thanks for the patch.

PR #125, opened 2021-05-22
Handle EINTR during epoll_wait.

### LINKED ISSUE #124: Signal handling for epoll_wait can leak processes
Unexpected signal delivery to the JVM while NuProcess's `ProcessEpoll` is blocked in `epoll_wait` can result in it leaking zombie processes.

Per `signal(7)`:
```
       The following interfaces are never restarted after being interrupted by
       a signal handler, regardless of the use of SA_RESTART; they always fail
       with the error EINTR when interrupted by a signal handler:

       * File     descriptor     multiplexing    interfaces:    epoll_wait(2),
         epoll_pwait(2), poll(2), ppoll(2), select(2), and pselect(2).
```

This means even if the JVM registers signal handlers with `SA_RESTART`, `epoll_wait` will not be restarted.

Looking through JDK's native code at places where it uses `epoll_wait`, it wraps them in a `RESTARTABLE` macro ([defined here](https://github.com/JetBrains/jdk8u_jdk/blob/94318f9185757cc33d2b8d527d36be26ac6b7582/src/solaris/native/sun/nio/ch/nio_util.h#L33-L37)) which detects a `-1` return paired with `EINTR` and re-runs `epoll_wait`. `ProcessEpoll` should have something similar.

Debugging this leak was quite challenging due to NuProcess's (lack of) error handling.
* `ProcessEpoll` [throws a `RuntimeException`](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java#L221) with a minimal message and no `errno` to indicate what the error was, but that doesn't really matter because
* `BaseEventProcessor.run` [catches the exception and aborts](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java#L84-L87) without any logging

I ended up [expanding the exception message](https://github.com/atlassian/NuProcess/commit/3ef3e06caf22c48dde81a11c4e2428cf2b0d8780#diff-4e5c48d2691d693151c8dde51faadc114e5d53003ecfe75b686736600398405fR236) and [adding some logging](https://github.com/atlassian/NuProcess/commit/3ef3e06caf22c48dde81a11c4e2428cf2b0d8780#diff-b466e66bd6d8b8c2049aebaef7e9a1a9a433ade7d5a37e89eef64f6dbd850161R94) to finally get to the bottom of it.

bturner:
`epoll_wait` is not restarted after it's interrupted by a signal handler, even if the handler used `SA_RESTART`. Instead, the calling code needs to handle the `EINTR` and restart the call itself.

- Added logging in `BaseEventProcessor` and `ProcessCompletions` when the `process()` loop is aborted due to an exception
- Added `EINTR` to `LibC`
- Updated `ProcessEpoll` to detect `EINTR` after `epoll_wait` fails and return `true` to retry the call

Fixes #124

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
@@ -21,6 +21,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import com.sun.jna.ptr.IntByReference;
 
@@ -33,6 +35,9 @@ public abstract class BaseEventProcessor<T extends BasePosixProcess> implements 
 
    protected static final int DEADPOOL_POLL_INTERVAL;
    protected static final int LINGER_ITERATIONS;
+
+   private static final Logger LOGGER = Logger.getLogger(BaseEventProcessor.class.getCanonicalName());
+
    private final int lingerIterations;
 
    protected Map<Integer, T> pidToProcessMap;
@@ -83,6 +88,8 @@ public abstract class BaseEventProcessor<T extends BasePosixProcess> implements 
       }
       catch (Exception e) {
          // TODO: how to handle this error?
+         LOGGER.log(Level.WARNING, "Aborting processing loop after unexpected exception (" +
+                 pidToProcessMap.size() + " processes running)", e);
          isRunning.set(false);
       }
       finally {
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -218,7 +218,17 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       try {
          int nev = LibEpoll.epoll_wait(epoll, triggeredEvent.getPointer(), 1, DEADPOOL_POLL_INTERVAL);
          if (nev == -1) {
-            throw new RuntimeException("Error waiting for epoll");
+            int errno = Native.getLastError();
+            if (errno == LibC.EINTR) {
+               // Signals received while in epoll_wait can interrupt the call and, per the documentation for
+               // SA_RESTART, it will not be restarted automatically. When that happens, we manually restart
+               // epoll_wait by returning true, which ensures BaseEventProcessor.run() calls process() again.
+               // This matches how the JVM internals handle EINTR during epoll_wait.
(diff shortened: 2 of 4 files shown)


src/main/java/com/zaxxer/nuprocess/windows/ProcessCompletions.java
```diff
@@ -122,7 +125,8 @@ public void run()
       }
       catch (Exception e) {
          // TODO: how to handle this error?
-         e.printStackTrace();
```
bturner, line 125: This is the last `e.printStackTrace` in the production NuProcess code; the last two are in tests (which seems fine)

src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
```diff
@@ -83,6 +88,8 @@ public void run()
       }
       catch (Exception e) {
          // TODO: how to handle this error?
+         LOGGER.log(Level.WARNING, "Aborting processing loop after unexpected exception (" +
+                 pidToProcessMap.size() + " processes running)", e);
```
bturner, line 92: This logging was instrumental to me in tracking down this issue. If there are other edge cases that cause the `run()` loop to die, this would make them a lot easier to solve going forward.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -218,7 +218,17 @@ public boolean process()
...
+            if (errno == LibC.EINTR) {
+               // Signals received while in epoll_wait can interrupt the call and, per the documentation for
+               // SA_RESTART, it will not be restarted automatically. When that happens, we manually restart
+               // epoll_wait by returning true, which ensures BaseEventProcessor.run() calls process() again.
+               // This matches how the JVM internals handle EINTR during epoll_wait.
+               // See https://github.com/JetBrains/jdk8u_jdk/blob/94318f9185757cc33d2b8d527d36be26ac6b7582/src/solaris/native/sun/nio/ch/nio_util.h#L33-L37
+               return true;
+            }
+
+            throw new RuntimeException("Error waiting for epoll (" + errno + ")");
```
bturner, line 231: Adding `errno` into the exception message here to aid debugging in case there are other errors. `epoll_wait` is documented as being able to trigger several others but, aside from `EINTR`, they all seem like the sorts of things where a blind retry wouldn't be the right handling.

bturner:
Seems like the CI bits aren't really running anymore. I don't see the ARM and x64 checks that other pull requests have had.

We were able to get a reproduction of `EINTR` by attaching `strace` to the JVM while it's in `epoll_wait`. Without this patch, an exception is thrown and the process is leaked. With it, processing continues normally.

bturner:
I see 2.0.2 was released without this. Any chance of a merge and a 2.0.3?

bturner:
We've found another path that can result in zombies. I'll either update this pull request or open a second one.

PR #133, opened 2021-09-30
Use Job objects in lieu of a shutdown hook on Windows.

doctorpangloss:
Shutdown hooks have bad behavior on Windows. Using Windows Job objects
ensure child processes die when the parent process (the NuProcess
caller) dies, regardless of the circumstances that caused the JVM
running NuProcess died.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
@@ -21,8 +21,6 @@ import com.sun.jna.Native;
 import com.sun.jna.WString;
 import com.zaxxer.nuprocess.NuProcess;
 import com.zaxxer.nuprocess.NuProcessHandler;
-import com.zaxxer.nuprocess.windows.NuKernel32.OVERLAPPED;
-import com.zaxxer.nuprocess.windows.NuWinNT.*;
 
 import java.nio.ByteBuffer;
 import java.nio.file.Path;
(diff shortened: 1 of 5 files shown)


src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -250,6 +250,53 @@ NuProcess start(List<String> commands, String[] environment, Path cwd)
...
    }
 
+   private void attachJob()
+   {
+      if (!CHILD_TERMINATES_WITH_PARENT) {
+         return;
+      }
+
+      if (!AssignProcessToJobObject(hJob, getPidHandle())) {
+         throw new RuntimeException("Failed to attach to job " + Native.getLastError());
```
bturner, line 260: You're almost certainly going to leak a process if this happens. `start` will catch this exception and (from my own changes) log it, and then return without the process ever being terminated.

src/main/java/com/zaxxer/nuprocess/windows/WindowsProcess.java
```diff
@@ -226,6 +222,8 @@ public void setProcessHandler(NuProcessHandler processHandler)
 
    NuProcess start(List<String> commands, String[] environment, Path cwd)
```
bturner, line 223: Shouldn't you also make changes to `run` below so it uses jobs too?

doctorpangloss:
you should probably vibe code the fix for this yourself, since it's a real issue

PR #137, opened 2022-04-21
Update various Maven details.

bturner:
- Added and configured `nexus-staging-maven-plugin` based on [current publishing instructions](https://central.sonatype.org/publish/publish-maven/#distribution-management-and-authentication)
- Added `mvnvm.properties`, which can be used with [mvnvm](https://mvnvm.org/) to indicate what Maven version to run
- Added explicit versions for relevant Maven plugins and updated all of the existing versions to current
- Moved configuration for the Javadoc and source plugins to the release profile
- Removed the `oss-parent` POM
  - [Current publishing instructions](https://central.sonatype.org/publish/publish-maven/#deprecated-oss-parent) for Central indicate the `oss-parent` POM is deprecated/unsupported and should not be used
- Ran `mvn tidy:pom` to apply best-practice ordering
- Renamed `release-sign-artifacts` profle to `release-profile`, matching the name of the default release profile
- Replaced `org.hamcrest:hamcrest-junit` with `hamcrest`, and updated to 2.2
- Updated JNA to 5.11, but modified the OSGi instructions to set the lower bound to 5.5 (matching the 2.0.0 release)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,11 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+        <!-- When releasing a new version, this property controls whether the staged artifacts will be automatically
+             released in Nexus. If this is set to false (-DautoReleaseStagedArtifacts=false), artifacts will need to
+             be manually released via the Nexus UI at https://oss.sonatype.org -->
+        <autoReleaseStagedArtifacts>true</autoReleaseStagedArtifacts>
     </properties>
 
     <dependencies>
@@ -249,6 +254,7 @@
                         <artifactId>nexus-staging-maven-plugin</artifactId>
                         <extensions>true</extensions>
                         <configuration>
+                            <autoReleaseAfterClose>${autoReleaseStagedArtifacts}</autoReleaseAfterClose>
                             <nexusUrl>https://oss.sonatype.org/</nexusUrl>
                             <serverId>ossrh</serverId>
                         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -2,8 +2,6 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <!-- For release: mvn release:perform -Darguments=-Dgpg.passphrase=PASSPHRASE -->
-
     <groupId>com.zaxxer</groupId>
     <artifactId>nuprocess</artifactId>
     <version>2.0.3-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


bturner:
I've done a trial 2.0.3-rc1 with these changes, and then dropped the staging repository and tag rather than publishing.

@brettwooldridge I'm going to merge this tomorrow so I can do a 2.0.3 release. If you have time before then to glance at it, great. If not, totally understandable. If there's anything you see here you don't like whenever you do have time to take a look, just add comments and I'll circle back to it in a follow-up pull request. (I've already taken way longer to get to where I think I can release 2.0.3 than I intended, so now I'm hoping to turn it around quickly.)

pom.xml
```diff
@@ -12,18 +10,10 @@
     <name>NuProcess</name>
     <description>A low-overhead, non-blocking I/O, external Process execution implementation for Java.</description>
     <url>https://github.com/brettwooldridge/NuProcess</url>
-
```
bturner, line 15: All of these ordering and whitespace changes are `mvn tidy:pom`

pom.xml
```diff
@@ -46,24 +36,35 @@
...
+            <email>incommand@hotmail.com</email>
+        </developer>
     </developers>
 
+    <scm>
+        <connection>scm:git:git@github.com:brettwooldridge/NuProcess.git</connection>
+        <developerConnection>scm:git:git@github.com:brettwooldridge/NuProcess.git</developerConnection>
+        <url>git@github.com:brettwooldridge/NuProcess.git</url>
+    </scm>
+    <distributionManagement>
```
bturner, line 50: This is part of the new setup for using `nexus-staging-maven-plugin` to do releases

pom.xml
```diff
@@ -46,24 +36,35 @@
...
+        </snapshotRepository>
+    </distributionManagement>
+
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
-    <parent>
-        <groupId>org.sonatype.oss</groupId>
-        <artifactId>oss-parent</artifactId>
```
bturner, line 57: Per the Central documentation, `oss-parent` is unmaintained and shouldn't be used

pom.xml
```diff
@@ -72,95 +73,133 @@
...
+            <artifactId>hamcrest</artifactId>
+            <version>2.2</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
 
     <build>
         <sourceDirectory>src/main/java</sourceDirectory>
         <testSourceDirectory>src/test/java</testSourceDirectory>
+        <pluginManagement>
```
bturner, line 85: All of these new plugin versions are current as of now. There are newer _milestones_ for some, but I've gone with release versions only.

pom.xml
```diff
@@ -72,95 +73,133 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <reuseForks>false</reuseForks>
                 </configuration>
-                <executions>
-                    <execution>
-                        <id>attach-sources</id>
-                        <goals>
-                            <goal>jar</goal>
```
bturner, line 133: Having these `jar` goals on the `maven-source-plugin` and `maven-javadoc-plugin` were resulting in sources and Javadocs being attached twice. I've confirmed via my trial 2.0.3-rc1 release that sources and Javadocs are still attached without these.

pom.xml
```diff
@@ -183,6 +223,36 @@
...
+                    <!-- nexus-staging-maven-plugin replaces the standard maven-deploy-plugin for releases and
+                         is useful for ensuring artifacts are all-or-nothing, as well as allowing artifacts to
+                         be reviewed before they're made public -->
+                    <plugin>
+                        <groupId>org.sonatype.plugins</groupId>
+                        <artifactId>nexus-staging-maven-plugin</artifactId>
+                        <extensions>true</extensions>
+                        <configuration>
+                            <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+                            <serverId>ossrh</serverId>
```
bturner, line 253: I have not added `<autoReleaseAfterClose>true</autoReleaseAfterClose>` here, which means after performing a release you have to either login to https://oss.sonatype.org and release the artifacts or use `mvn nexus-staging:release`. Not using auto-release was useful in testing my release setup, but it may be desirable to set it here going forward.

bturner:
I've rebased my pull request to sign all of my commits. I've also added a property to allow specifying whether to auto-release staged artifacts or not (defaulting to `true`, though I'll probably override it to `false` when I release 2.0.3 just to be safe)

PR #138, opened 2022-04-23
Remove static EpollEvent queue. (fixes #126)

### LINKED ISSUE #126: Interrupts while registering processes with ProcessEpoll can leak processes
`ProcessEpoll` maintains a `BlockingQueue` of `EpollEvent` instances. When processes are registered, an instance is pulled off the queue, used to register events for that process, and then put back onto the queue. The queue has 64 `EpollEvent` instances in it, so In a system that's lightly loaded there's always an instance available. In a heavily loaded Bitbucket Server instance, though, spikes can easily trigger the creation of 100+ processes concurrently.

If the thread is interrupted (e.g. `Thread.interrupt()`) while it's blocked in `take()`, a `RuntimeException` is thrown and caught by `LinuxProcess`'s `start` or `run` code, which logs the exception with a message that the process couldn't be started and the process is abandoned. But that's not actually accurate. The process _was_ started, but NuProcess couldn't register it. That means NuProcess never reaps the process and it either hangs (if it needs `stdin` or blocks after writing too much to `stdout` or `stderr`) or terminates and becomes a zombie.

bturner:
When creating many processes concurrently under load, it's possible to exhaust the 64 `EpollEvent`s in the queue and end up with threads blocked waiting for one to free up. If such a thread gets interrupted, it ends up leaking the process as a zombie.

I ran the jmh microbenchmark before and after this change to check the allocation rate and found no measurable difference. There was also no measurable difference in performance.

- Added an `EpollEvent` field to `LinuxProcess`
  - This way each process has its own event instance that can be used (and reused) whenever the processing thread needs to reconfigure the `epoll` registration
  - Since each write to `stdin` requires its own `EPOLLONESHOT` event, having a reusable `EpollEvent` avoids allocation churn
- Removed the `eventPool` in `ProcessEpoll` and updated it to use the process's `EpollEvent` when necessary
- Updated wording for a few `RuntimeException`s to make their messages a little more consistent, and added `errno` values to some that weren't including them to assist with debugging/analysis

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -19,8 +19,6 @@ package com.zaxxer.nuprocess.linux;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
 
 import com.sun.jna.Native;
 import com.sun.jna.ptr.IntByReference;
@@ -38,22 +36,11 @@ import static com.zaxxer.nuprocess.internal.LibC.WTERMSIG;
  */
 class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
 {
-   private static final int EVENT_POOL_SIZE = 64;
-   private static final BlockingQueue<EpollEvent> eventPool;
+   private final int epoll;
+   private final EpollEvent triggeredEvent;
+   private final List<LinuxProcess> deadPool;
 
-   private int epoll;
-   private EpollEvent triggeredEvent;
-   private List<LinuxProcess> deadPool;
    private LinuxProcess process;
-
-   static
-   {
-      eventPool = new ArrayBlockingQueue<>(EVENT_POOL_SIZE);
-      for (int i = 0; i < EVENT_POOL_SIZE; i++) {
-         EpollEvent event = new EpollEvent();
-         eventPool.add(event);
-      }
-   }
 
    ProcessEpoll()
    {
@@ -76,7 +63,8 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
 
       epoll = LibEpoll.epoll_create(1024);
       if (epoll < 0) {
-         throw new RuntimeException("Unable to create kqueue: " + Native.getLastError());
+         int errno = Native.getLastError();
+         throw new RuntimeException("Unable to create kqueue, errno: " + errno);
       }
 
       triggeredEvent = new EpollEvent();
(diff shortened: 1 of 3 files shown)


src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -160,22 +138,19 @@ public void queueWrite(LinuxProcess process)
...
+         EpollEvent event = process.getEpollEvent();
          event.setEvents(LibEpoll.EPOLLOUT | LibEpoll.EPOLLONESHOT | LibEpoll.EPOLLRDHUP | LibEpoll.EPOLLHUP);
          event.setFileDescriptor(stdin);
          int rc = LibEpoll.epoll_ctl(epoll, LibEpoll.EPOLL_CTL_MOD, stdin, event.getPointer());
          if (rc == -1) {
             LibEpoll.epoll_ctl(epoll, LibEpoll.EPOLL_CTL_DEL, stdin, event.getPointer());
             rc = LibEpoll.epoll_ctl(epoll, LibEpoll.EPOLL_CTL_ADD, stdin, event.getPointer());
+            if (rc == -1) {
+               int errno = Native.getLastError();
+               throw new RuntimeException("Unable to register new event to epoll queue, errno: " + errno);
```
bturner, line 151: This change to add `errno` values stems from some debugging issues we had with errors during `closeStdin` calls in Bitbucket Server. If a process racily terminates while we're trying to close `stdin`, this exception ends up thrown with `EBADF` (9). I'll try to work on a separate patch that will actually handle that case cleanly, but for the moment this at least makes it possible to identify the case where closing `stdin` raced with process termination.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -38,23 +36,12 @@
...
+   private final List<LinuxProcess> deadPool;
 
-   private int epoll;
-   private EpollEvent triggeredEvent;
-   private List<LinuxProcess> deadPool;
    private LinuxProcess process;
 
-   static
-   {
-      eventPool = new ArrayBlockingQueue<>(EVENT_POOL_SIZE);
```
bturner, line 51: Bitbucket Server under load can be running hundreds of processes concurrently and, since it generally uses `NuProcess.run` instead of `NuProcess.start`, it can be pumping those processes on hundreds of threads. Under those conditions, it's easy for the system to run out of events in the queue and block waiting for one to be returned. If a blocked thread got interrupted (e.g. because a request was canceled), the exception thrown would end up causing NuProcess to abandon the process, resulting in a zombie.

This static queue was added to combat memory churn (see 9df9b4ccdc055662c0406890fb1e419211ca694d), but in practice there's no observable difference in churn (at least per the jmh microbenchmark that was used to analyze the change above when it was made) between having a reusable `EpollEvent` on each `LinuxProcess` and using this static queue. Given that, this approach of an event per process is simpler and avoids any contested resources.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -228,7 +203,7 @@ public boolean process()
                return true;
             }
 
-            throw new RuntimeException("Error waiting for epoll (" + errno + ")");
+            throw new RuntimeException("Error waiting for epoll, errno: " + errno);
```
bturner, line 206: Just rewording this to make the way `errno` is communicated more consistent across errors

PR #151, opened 2024-02-11
Don't depend on the GC to free native memory on Linux

avrecko:
JNA is freeing 192 KiB of per NuProcess native buffers on GC Finalizer / Reference Queue processing. In cases with very low GC activity, this is essentially a memory leak, potentially causing native memory issues.

  * out/err/in buffers are freed onExit,
  * event needed for Epoll #registerProcess and #queueWrite is allocated and freed in the method,
  * tweak in reusing of IntByReference for duration of epoll/kqueue processor,
  * updated JNA library to 5.13.0, to get #close on Memory object (added in 5.12.0).

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
@@ -48,6 +48,8 @@ public abstract class BaseEventProcessor<T extends BasePosixProcess> implements 
    private CyclicBarrier startBarrier;
    private AtomicBoolean isRunning;
 
+   protected final IntByReference exitCodePointer = new IntByReference();
+
    static {
       LINGER_TIME_MS = Math.max(1000, Integer.getInteger("com.zaxxer.nuprocess.lingerTimeMs", 2500));
 
@@ -128,6 +130,7 @@ public abstract class BaseEventProcessor<T extends BasePosixProcess> implements 
          process.onExit(Integer.MAX_VALUE - 1);
          LibC.waitpid(process.getPid(), exitCode, LibC.WNOHANG);
       }
+      Util.close(exitCode);
    }
 
    /**
--- a/src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import com.sun.jna.*;
+import com.zaxxer.nuprocess.internal.Util;
 
 class EpollEvent
 {
@@ -62,6 +63,11 @@ class EpollEvent
 
    int size() {
       return size;
+   }
+
+   public void close()
+   {
+      Util.close(getPointer());
    }
 
    public static class EpollEventPrototype extends Structure
--- a/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
@@ -21,7 +21,6 @@ import java.util.LinkedList;
 import java.util.List;
 
 import com.sun.jna.Native;
-import com.sun.jna.ptr.IntByReference;
 import com.zaxxer.nuprocess.NuProcess;
 import com.zaxxer.nuprocess.internal.BaseEventProcessor;
 import com.zaxxer.nuprocess.internal.LibC;
@@ -86,6 +85,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
       int stdinFd = Integer.MIN_VALUE;
       int stdoutFd = Integer.MIN_VALUE;
       int stderrFd = Integer.MIN_VALUE;
+      EpollEvent event = new EpollEvent();
       try {
          stdinFd = process.getStdin().acquire();
          stdoutFd = process.getStdout().acquire();
@@ -96,7 +96,6 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          fildesToProcessMap.put(stdoutFd, process);
          fildesToProcessMap.put(stderrFd, process);
 
-         EpollEvent event = process.getEpollEvent();
          event.setEvents(LibEpoll.EPOLLIN);
          event.setFileDescriptor(stdoutFd);
          int rc = LibEpoll.epoll_ctl(epoll, LibEpoll.EPOLL_CTL_ADD, stdoutFd, event.getPointer());
@@ -114,6 +113,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          }
       }
       finally {
+         event.close();
          if (stdinFd != Integer.MIN_VALUE) {
             process.getStdin().release();
          }
@@ -133,13 +133,14 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          return;
       }
 
+      EpollEvent event = null;
       try {
          int stdin = process.getStdin().acquire();
          if (stdin == -1) {
            return;
          }
 
-         EpollEvent event = process.getEpollEvent();
+         event = new EpollEvent();
          event.setEvents(LibEpoll.EPOLLOUT | LibEpoll.EPOLLONESHOT | LibEpoll.EPOLLRDHUP | LibEpoll.EPOLLHUP);
          event.setFileDescriptor(stdin);
          int rc = LibEpoll.epoll_ctl(epoll, LibEpoll.EPOLL_CTL_MOD, stdin, event.getPointer());
@@ -153,6 +154,9 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          }
       }
       finally {
+         if (event != null) {
+            event.close();
+         }
          process.getStdin().release();
       }
    }
@@ -306,8 +310,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          return;
       }
 
-      IntByReference ret = new IntByReference();
-      int rc = LibC.waitpid(linuxProcess.getPid(), ret, LibC.WNOHANG);
+      int rc = LibC.waitpid(linuxProcess.getPid(), exitCodePointer, LibC.WNOHANG);
 
       if (rc == 0) {
          deadPool.add(linuxProcess);
@@ -316,7 +319,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          linuxProcess.onExit((Native.getLastError() == LibC.ECHILD) ? Integer.MAX_VALUE : Integer.MIN_VALUE);
       }
       else {
-         handleExit(linuxProcess, ret.getValue());
+         handleExit(linuxProcess, exitCodePointer.getValue());
       }
    }
 
@@ -326,11 +329,10 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
          return;
       }
 
-      IntByReference ret = new IntByReference();
       Iterator<LinuxProcess> iterator = deadPool.iterator();
       while (iterator.hasNext()) {
          LinuxProcess process = iterator.next();
-         int rc = LibC.waitpid(process.getPid(), ret, LibC.WNOHANG);
+         int rc = LibC.waitpid(process.getPid(), exitCodePointer, LibC.WNOHANG);
          if (rc == 0) {
             continue;
          }
@@ -341,7 +343,7 @@ class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
             continue;
          }
 
-         handleExit(process, ret.getValue());
+         handleExit(process, exitCodePointer.getValue());
       }
    }
 
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <dependency>
             <groupId>net.java.dev.jna</groupId>
             <artifactId>jna</artifactId>
-            <version>5.11.0</version>
+            <version>5.13.0</version>
         </dependency>
 
         <dependency>
--- a/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/BasePosixProcess.java
@@ -323,8 +323,9 @@ public abstract class BasePosixProcess implements NuProcess
       }
       finally {
          exitPending.countDown();
-         // Once the last reference to the buffer is gone, Java will finalize the buffer
-         // and release the native memory we allocated in initializeBuffers().
+         outBufferMemory.close();
+         errBufferMemory.close();
+         inBufferMemory.close();
          outBufferMemory = null;
          errBufferMemory = null;
          inBufferMemory = null;
@@ -332,7 +333,6 @@ public abstract class BasePosixProcess implements NuProcess
          errBuffer = null;
          inBuffer = null;
          processHandler = null;
-         Memory.purge();
       }
    }
 
--- a/src/main/java/com/zaxxer/nuprocess/internal/Util.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/Util.java
@@ -1,0 +1,22 @@
+package com.zaxxer.nuprocess.internal;
+
+import com.sun.jna.Memory;
+import com.sun.jna.Pointer;
+import com.sun.jna.PointerType;
+
+public class Util {
+
+    public static void close(PointerType pointerType) {
+        if (pointerType == null) {
+            return;
+        }
+        close(pointerType.getPointer());
+    }
+
+    public static void close(Pointer p) {
+        if (p instanceof Memory) {
+            ((Memory) p).close();
+        }
+    }
+
+}
--- a/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
@@ -35,7 +35,6 @@ import static com.zaxxer.nuprocess.internal.Constants.JVM_MAJOR_VERSION;
  */
 public class LinuxProcess extends BasePosixProcess
 {
-   private final EpollEvent epollEvent;
 
    static {
       LibEpoll.sigignore(LibEpoll.SIGPIPE);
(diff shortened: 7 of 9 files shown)


src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -36,10 +39,19 @@
  */
 class ProcessEpoll extends BaseEventProcessor<LinuxProcess>
 {
+   private static final int EPOLL_MAX_EVENTS = 20;
    private final int epoll;
-   private final EpollEvent triggeredEvent;
+   private final EpollEvent[] triggeredEvents;
+   private final EpollEvent firstTriggeredEvent;
```
bturner, line 45: There's a fair bit of history around the epoll code in pull request #80 that might be worth checking, as well as on commit 9df9b4ccdc055662c0406890fb1e419211ca694d (which were optimizations Brett did on top of mine).

One thing I didn't discuss on that pull request was that I profiled the difference between passing a single `EpollEvent` to `epoll_wait`  and passing an array, and I found no statistically significant difference between the two. (I wish I still had access to my write-up on the subject, but I lost that when I left Atlassian.)

I'm curious if there's something specific you're trying to accomplish with this change--and whether you have any measurements that _show_ it being accomplished. Because my own testing on this, in the past, showed there was no benefit, even under heavy process concurrency.

Ultimately, though, that's not really my primary issue with this. The real problem is you haven't accounted for `LinuxProcess.run`, only `LinuxProcess.start`. When using `start`, NuProcess's pool of long-lived `ProcessEpoll` threads handle everything. With `run`, though, _each `run` call_ uses its own short-lived `ProcessEpoll`. With this change, that means all of those `run` calls (which, again, Bitbucket Server uses _extensively_--I'd say 80-90% of its processes use `run` rather than `start`) is now allocating 20 `EpollEvent`s despite having no chance of ever needing more than 3 (in the absolute worst case; in practice I suspect it'd rarely use more than 1).

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -326,7 +367,7 @@ private void checkDeadPool()
          return;
       }
 
-      IntByReference ret = new IntByReference();
+      IntByReference ret = tempIntPointer;
```
bturner, line 370: I'd drop the alias here and just update usages to refer to the field directly.

src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
```diff
@@ -79,7 +83,7 @@ public static class EpollEventPrototype extends Structure
       public int events;
       public EpollData data;
 
-      EpollEventPrototype() {
+      public EpollEventPrototype() {
```
bturner, line 86: Why did this need to be made `public`? (That said, though, in light of other comments I've added, I think you can just revert this file entirely.)

src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
```diff
@@ -48,6 +48,10 @@ public abstract class BaseEventProcessor<T extends BasePosixProcess> implements
    private CyclicBarrier startBarrier;
    private AtomicBoolean isRunning;
 
+   // avoid unnecessary malloc calls by reusing native pointer
+   // this field is thread safe as this is only used by Epoll/KQueue thread
+   protected final IntByReference tempIntPointer = new IntByReference();
```
bturner, line 53: I don't like `temp` names for things that aren't temporary. This doesn't get allocated and then cleaned up; it gets reused over and over. I'd call it something like `exitCodePointer` or something like that.

bturner (changes requested):
I appreciate your time and effort, @avrecko, but there are some parts of this I can't accept. I'd like to see it stripped back to a change that's more targeted to freeing the buffers, without some of the other changes, before I could accept it.

If you want to move forward with upgrading JNA to 5.13, using a single buffer allocation, adding an explicit `free` for it, and potentially keeping your `IntByReference` tweak, those changes can be pretty readily accepted. The rest, while perhaps beneficial to certain use cases, penalizes others.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -67,7 +79,17 @@ private ProcessEpoll(int lingerIterations)
...
-      triggeredEvent = new EpollEvent();
+      triggeredEvents = new EpollEvent[EPOLL_MAX_EVENTS];
+
+      EpollEvent.EpollEventPrototype epollEvent = Structure.newInstance(EpollEvent.EpollEventPrototype.class);
+      EpollEvent.EpollEventPrototype[] array = (EpollEvent.EpollEventPrototype[]) epollEvent.toArray(EPOLL_MAX_EVENTS);
+
+      for (int i = 0; i < EPOLL_MAX_EVENTS; i++) {
+         EpollEvent.EpollEventPrototype proto = array[i];
+         triggeredEvents[i] = new EpollEvent(proto.getPointer());
+      }
```
bturner, line 90: I'm not sure I really understand the benefit here. The existing code (without your changes) allocates exactly one `EpollEventPrototype` ever, in the `static` initializer for `EpollEvent`, to determine the layout/size of the structure. Everything past that allocates `EpollEvent` directly--without a prototype.

Here, you've replaced the `EpollEvent` creation/initialization with creation of an array of prototypes, and then you reuse their memory by re-wrapping them as `EpollEvent`. How is that "better" than just using `EpollEvent` directly?
```java
EpollEvent epollEvent = Strucutre.newInstance(EpollEvent.class);
EpollEvent[] array = (EpollEvent[]) epollEvent.toArray(EPOLL_MAX_EVENTS);
```

In both cases, you're only allocating the underlying JNA memory once. Did you think the current code allocated it twice? I suspect your changes here actually yield _more_ overall allocations per `EpollEvent`, when you consider both the native side and the JVM side, because `EpollEventPrototype` has its own `EpollData` nested union it creates, and then you _still_ need to create an `EpollEvent` too.

README.md
```diff
@@ -177,7 +177,7 @@ Julien Viet has developed a Vert.x 3 library [here](https://github.com/vietj/ver
 #### Limitations
 The following limitations exist in NuProcess:
  * Currently only supports Linux, Windows, and MacOS X.
- * Java 7 and above
+ * Java 8 and above
```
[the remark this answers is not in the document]
avrecko: Ok, works for me. 5.13.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
```diff
@@ -36,10 +39,19 @@
...
+   private final EpollEvent[] triggeredEvents;
+   private final EpollEvent firstTriggeredEvent;
    private final List<LinuxProcess> deadPool;
 
+   // preferring (assuming uncontented) locking over memory allocations
+   // used in #registration and #queueWrite methods
+   private final EpollEvent tempEventForRegistration = new EpollEvent();
+   private final EpollEvent tempEventForQueueWrite = new EpollEvent();
+   private final Lock tempEventForRegistrationLock = new ReentrantLock();
+   private final Lock tempEventForQueueWriteLock = new ReentrantLock();
```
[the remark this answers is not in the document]
avrecko: Hm. I put in the comment that I prefer _assumed_ uncontented locking over jna native call to malloc. Indeed I am optimizing for my use case at the cost of yours.

Since we are going to malloc for the buffers per process anyway might as well for event per process.

src/main/java/com/zaxxer/nuprocess/linux/ProcessEpoll.java
avrecko: I'll get back to you on this. In my unit test I already saw > 1 events being handled per one epoll select. I did not look into run.

In case of run(). I did just skim over the code. I do wonder. Did you ever try having a pool of selectors/epoll for the run use case? Instead of creating per process selector?

src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
avrecko: It is for using the JNA Struct API. It requires public constructor.

I am scratching my head a bit, why are there both the JNA Struct (EpollEventPrototype) and the POJO EpollEvent for the same thing? Hm.

I don't mind the extra few java objects. But we can do better.

avrecko:
@bturner What are your thoughts on explicitly freeing the per process EpollEvent on call to onExit()? The complication is that some synchronization guards are needed to avoid any API misuse that can cause SegFault.

E.g. `process.start().waitFor(0,Days);` then doing `process.wantWrite();` can cause SegFault in rare case (easy to reproduce) if EpollEvent native memory is freed onExit and no additional guards are put in place in queueWrite.

avrecko:
Also what are you thoughts of making the buffer configurable? Like a global setting?

src/main/java/com/zaxxer/nuprocess/linux/EpollEvent.java
bturner: Having two objects is part of "doing better". JNA `Structure`s are expensive to allocate, in a number of ways. The approach taken here is to instantiate the structure _once_ and use that to understand the memory layout. Then, going forward, `EpollEvent` is a simple POJO which allocates a simple `Memory` region of the right size. `EpollEvent` has no fields as such; it just writes into the memory region at the right locations for the fields the structure had.

I linked to the commit where that optimization was made, and in pull request #80 there are comments which show memory churn before and after that change--showing it's a pretty significant win.

bturner:
> What are your thoughts on explicitly freeing the per process EpollEvent on call to onExit()?

For this review, I'd leave it out. Synchronization brings in its own costs that need to be balanced. I'm not saying we shouldn't do it, but rather that it's better to separate "contentious" changes from the simple ones that can be readily merged. That way I can accept this pull request and do a new release with some improvements, and then we can have a more focused discussion about further improvements and the trade-offs they entail.

> Also what are you thoughts of making the buffer configurable? Like a global setting?

I'm not sure what buffer you're referring to. The size for the `stderr`, `stdout` and `stdin` buffers? Something else?

avrecko:
Ok agreed. I'll put the explicit close of per process EpollEvent in another
pull request.

Yeah, I saw the offset handling regarding the EPoll struct. Wow. Didn't
image JNA to be this inefficient to merit doing it yourself with offsets.
Not very convenient then.

fwiw: doing some simple testing also not seeing the benefit of polling 20
instead of 1 event. I do consistently see 2 events queried. But anyway, for
now I left the poll alone.

For the configuration I meant for setting of NuProcess#BUFFER_CAPACITY. For
my use case, I think 4K would work just fine.

I'll clean up the code and let you know when it is ready to be reviewed
again.

V V pon., 12. feb. 2024 ob 19:23 je oseba Bryan Turner <
***@***.***> napisala:

bturner:
> fwiw: doing some simple testing also not seeing the benefit of polling 20
> instead of 1 event. I do consistently see 2 events queried. But anyway, for
> now I left the poll alone.

That matches what I saw before. I could see >1 event being returned, but performance and profiling characteristics were identical either way. (Polling for more can reduce the number of syscalls, but since handler processing dominates the performance so strongly, taking away milliseconds (or, more likely, nanoseconds) of syscall execution simply doesn't move the needle.) Based on that, I think leaving it at single poll is probably fine (and keeps the code simpler).

> For the configuration I meant for setting of NuProcess#BUFFER_CAPACITY. For
> my use case, I think 4K would work just fine.

Brett and I had previously at least touched on the idea of making that configurable ([see here](https://github.com/brettwooldridge/NuProcess/pull/80#discussion_r161377720)). A system property approach would be pretty straightforward, so it might be a good place to start. I've thought before about what might be involved in making it more readily runtime-configurable (e.g. so that, when you start a process, you can request a buffer size as part of that, because not all processes have the same I/O needs), but haven't actually sat down to try and do it.

avrecko:
Have same thoughts about setting of out/err/in buffer sizes. I'll prepare
another pull request for system property approach. I also think it is a
good place to start. For more fine grained control, process builder is
probably the place to add it. But maybe unnecessary.

avrecko:
@bturner I think this is it. Updated JNA to get the #close method, out/err/in buffers are closed and added the tweak to reuse IntByReference pointer by the epoll/kqueue.

Btw: wow, the Kqueue is using the Struct KEvent, indeed it is heavy. I expected some bespoke code generators from JNA. Not even close.

I still have my eyes on the per process Event allocation that isn't freed right away. I'll think of something.

avrecko:
For the per process event. I know you'd prefer to do this in a different PR. But I think I came up with a solution that is simple enough.

Event is needed by the `registerProcess()` and by the `wantWrite()` and `writeStdin(ByteBuffer)`.

For the registration. We can allocate the event and free right away inside the registration method. This way the per process event is not needed.

For stdin handling. I prefer on demand init of the event that is then used by calling wantWrite and writeStdin. Technically multiple threads can call wantWrite and writeStdin. But the race is not an issue since all threads are setting the same values that are atomic (on x86 at least).

```
      event.setEvents(LibEpoll.EPOLLOUT | LibEpoll.EPOLLONESHOT | LibEpoll.EPOLLRDHUP | LibEpoll.EPOLLHUP);
      event.setFileDescriptor(stdin);
```

If 2 threads would be calling `writeStdin`  and the fields would be 64bits, a tear could occur. But this is effectively tread safe (on platforms where 32bit are atomic). I see you use `ConcurrentLinkedQueue<ByteBuffer>` so I assume concurrent access to `writeStdin` is expected.

Apart from on demand per process init of Event when doing `wantWrite` and `writeStdin`. I also added `closeStdinEpollEvent`. This way the control of Event is on the caller.

Let me know what you think. Like mentioned I want all of the malloc stuff to be explicity freed when process exits. With this changes this is achieved for `start()`.

bturner:
That approach doesn't really make sense to me, unfortunately.
* It requires _more_ allocations, since the `EpollEvent` for registration is no longer reused for `wantWrite`; anything that will write will now result in creation of (minimum) 2 `EpollEvent`s where the existing code would allocate 1
* It leaks the implementation details above the API, because the caller is now made aware that a) there's an `EpollEvent` that b) needs to be manually released

For point 1, if we're not massively concerned about allocating a few `EpollEvent`s (and, so long as they're explicitly released, I'm not sure we _need_ to be), then I'm not sure why you wouldn't make a similar change in `ProcessEpoll::queueWrite` to what you have made in `registerProcess` and "just" have `queueWrite` create a short-lived `EpollEvent`, use it to register and then `close` it.

For point 2, if we don't want to allocate an `EpollEvent` in each `queueWrite` call, I would revert your changes to `registerProcess`, so it uses the single `EpollEvent`, which would continue to be allocated during `LinuxProcess` creation, for both registration and writes, and I'd replace `closeStdinEpollEvent` with a more generic `releaseNative` or `cleanup`. This leads to no more allocations than the code currently has, while still giving callers a mechanism to "force" cleanup rather than waiting.

If we're going to add a new method, I don't think it's "correct" to add it just to `LinuxProcess`. Callers shouldn't be casting to the specific process classes; they should be using `NuProcess`. Based on that, I'd add whatever method we add for this to that interface. (It can't have a `default` implementation since we still support Java 7, but the existing classes can add no-op implementations.)

Personally, at this point, I'm inclined to go with "just" allocating `EpollEvent`s in `queueWrite` and closing them, or just leave the code as-is. Adding new "cleanup" handling through `NuProcess` feels like a really heavy solution to a "problem" I'm still not sure I understand. (Read: Optimizing for this feels like a bit of a waste of time, to me. JNA _will_ free the native memory; it's just not the sort of deterministic schedule you personally would like it to be on. At 12 bytes per `EpollEvent`, you'd need to have allocated tens of thousands of them for the memory usage to add up to anything significant--and you'd need JNA to have not freed any of them on top of that. In multiple years of working on Bitbucket Server with it using NuProcess under the hood, we never once experienced internally or received a report externally of an instance of an OOM error related to JNA's native memory handling--and that's with the 192KB of buffers not being explicitly freed on top of the 12 bytes for an `EpollEvent`, despite forking millions of `git` processes over the JVM's lifetime.)

src/main/java/com/zaxxer/nuprocess/internal/BaseEventProcessor.java
```diff
@@ -128,6 +131,7 @@ public void shutdown()
          process.onExit(Integer.MAX_VALUE - 1);
          LibC.waitpid(process.getPid(), exitCode, LibC.WNOHANG);
       }
+      ((Memory) exitCode.getPointer()).close();
```
bturner, line 134: I don't really like this assumption about the internals of JNA. You're downcasting to a derived type (`Memory` extends `Pointer`), which, _per the API_, may fail, and you're doing it in an unchecked way (as in, if the cast fails, you're not handling it)--all to free 4 bytes of memory that will end up automatically freed at some point without your intervention. Especially for the set of `ProcessEpoll` threads that handle `start`ed processes, which only terminate when the JVM does, the benefit of this cross-domain trespass feels effectively nonexistent.

Given `Memory` implements `Closeable`, if we want to do this, I wonder if we should consider adding a simple helper method somehwere in this `com.zaxxer.nuprocess.internal` package along the lines of:
```java
public static void closeIfCloseable(Object object) {
    if (object instanceof Closeable) {
        ((Closeable) object).close();
    }
}
```

Then all the different places you do this cast (and there look to be several), you can use the helper instead.

avrecko: Spot on. I didn't like it either. I just added a generic sounding Util method with close methods.

avrecko:
@bturner If you are cool with allocating/freeing per `queueWrite` this simplifies things greatly.

I think JNA and also JNI does some small allocations per method call anyway. So probably not a big deal if we do as well.

avrecko:
I've updated the title to be less generic and update the first comment message with the details.

I've updated the code to reflect the latest discussion.

avrecko:
@bturner I left the `run` use case alone. Could also add explicit free for `run` use case. But would have to add a method probably `releaseNative` to the processor. I actually added this for Linux but then it makes more sense to add it to the top level and implement for all of the 3 OSes. So I reverted the change.

Leaving `run` use case as is. Allocate and free per method for register and queueWrite works for me. We can leave freeing per processor jna fields alone as they don't cause issues with GC for `start` use case. They are per `classloader` for `start` use case. This works fine for me.

To iterate the problem I am solving. In some GC setups. It might take hours between GC cycles that trigger reference processing. In this time nonnegligible amouth of native memory accumulate that cause native memory issues.

See: https://mail.openjdk.org/pipermail/zgc-dev/2023-August/001260.html

PR #152, opened 2024-02-13
Support for configuring NuProcess buffer capacity via user property

avrecko:
Configurable via property `com.zaxxer.nuprocess.bufferCapacity`.

I've set the limits between 1 KiB and 1 MiB. They are arbitrary choices. Should handle use cases of 4, 8, 16, 32, 64 KB/KiB.

edit: could also set limit to something like 128 KiB? Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,10 @@ This property controls how long the processing thread(s) remains after the last 
 order to avoid the overhead of starting up another processing thread, if processes are frequently run it may be desirable
 for the processing thread to remain (linger) for some amount of time (default 2500ms).
 
+##### ``com.zaxxer.nuprocess.bufferCapacity``
+Sets the size of native `ByteBuffer` used for `NuProcessHandler#onStdout`, `#onStderr` and `#onStdinReady` methods.
+The default value is 65536 Bytes (64 KiB). Minimum value is 1 KiB and maximum 1 MiB. 
+
 #### Related Projects
 Charles Duffy has developed a Clojure wrapper library [here](https://github.com/threatgrid/asynp).
 Julien Viet has developed a Vert.x 3 library [here](https://github.com/vietj/vertx-childprocess).
--- a/src/main/java/com/zaxxer/nuprocess/NuProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/NuProcess.java
@@ -16,6 +16,8 @@
 
 package com.zaxxer.nuprocess;
 
+import com.zaxxer.nuprocess.internal.Constants;
+
 import java.nio.ByteBuffer;
 import java.util.concurrent.TimeUnit;
 
@@ -40,7 +42,9 @@ import java.util.concurrent.TimeUnit;
  */
 public interface NuProcess
 {
-   int BUFFER_CAPACITY = 65536;
+   String BUFFER_CAPACITY_PROPERTY = "com.zaxxer.nuprocess.bufferCapacity";
+
+   int BUFFER_CAPACITY = Constants.getBufferCapacity();
 
    /**
     * Waits for the process to exit in a blocking fashion. See
--- a/src/main/java/com/zaxxer/nuprocess/internal/Constants.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/Constants.java
@@ -1,10 +1,21 @@
 package com.zaxxer.nuprocess.internal;
 
+import com.zaxxer.nuprocess.NuProcess;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 public class Constants
 {
+   private static final Logger LOGGER = Logger.getLogger(Constants.class.getCanonicalName());
+
    public static final int NUMBER_OF_THREADS;
    public static final int JVM_MAJOR_VERSION;
    static final OperatingSystem OS;
+
+   static final int DEFAULT_BUFFER_CAPACITY = 64 * 1024;
+   static final int MIN_BUFFER_CAPACITY = 1024;
+   static final int MAX_BUFFER_CAPACITY = 1024 * 1024;
 
    enum OperatingSystem
    {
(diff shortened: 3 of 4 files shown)


src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
```diff
@@ -310,4 +310,26 @@ private String[] prepareEnvironment()
 
       return env;
    }
+
+   static int getGlobalBufferSizeSetting()
```
bturner, line 314: I don't think this should be here. I'd put it on `NuProcess` instead. If a point comes where the `NuProcessBuilder` grows the ability to control the buffer size per-process, `NuProcess.BUFFER_CAPACITY` would still not be able to reflect that (since it's a static constant). I suspect if that ever became possible we'd need to deprecate that constant and either remove it or rename it (e.g. `DEFAULT_BUFFER_CAPACITY`). Regardless, though, it still wouldn't need to look at the builder to know what that value is.

src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
```diff
@@ -310,4 +310,26 @@ private String[] prepareEnvironment()
...
+         {
+            throw new IllegalArgumentException("This -D" + NuProcess.BUFFER_CAPACITY_PROPERTY + "=" + value + " needs to be between 1 KiB and 1 MiB.");
+         }
+         return value;
+      }
+      catch (NumberFormatException e)
+      {
+         throw new IllegalArgumentException("This -D" + NuProcess.BUFFER_CAPACITY_PROPERTY + "=" + bufferCapacityProperty + " needs to be a number.");
+      }
+   }
```
bturner, line 334: Please follow the code conventions of this project.
* Opening brace on same line
* For subsequent keywords (like `catch` or `else`) are on a separate line from the closing brace (as you have it)
* No unnecessary `final` qualifiers

Something like this:
```java
   static int getGlobalBufferSizeSetting() {
      String capacity = System.getProperty(NuProcess.BUFFER_CAPACITY_PROPERTY);
      if (capacity == null || capacity.trim().isEmpty()) {
         return 64 * 1024;
      }
      try {
         int value = Integer.parseInt(capacity);
         if (value < 1024) {
             // Log with JUL
             return 1024;
         }
         if (value > 1024 * 1024) {
             // Log with JUL
             return 1024 * 1024;
         }
         return value;
      }
      catch (NumberFormatException e) {
         // Log with JUL
         return 64 * 1024;
      }
   }
```

bturner:
Re: using 128K as the limit, I don't thing we need to super prescriptive here in the library. To some extent, I'm a bit inclined to just let the caller use whatever value they want, even if it's what _I_ would consider excessively small or large, and not validate at all beyond ensuring it's a number (and defaulting the value if it isn't, rather than throwing).

If we're going to have a limit, I'd probably leave it as you have it.

src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
```diff
@@ -310,4 +310,26 @@ private String[] prepareEnvironment()
...
+      if (bufferCapacityProperty == null || bufferCapacityProperty.trim().isEmpty())
+      {
+         return 64 * 1024;
+      }
+      try
+      {
+         final int value = Integer.parseInt(bufferCapacityProperty);
+         if (value < 1024 || value > 1024 * 1024)
+         {
+            throw new IllegalArgumentException("This -D" + NuProcess.BUFFER_CAPACITY_PROPERTY + "=" + value + " needs to be between 1 KiB and 1 MiB.");
```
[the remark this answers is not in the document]
avrecko: Totally fine with me to do it this way. I prefer to not allow invalid configuration.

src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
avrecko: Yeah, maybe we put this in Constants class?

avrecko:
> Re: using 128K as the limit, I don't thing we need to super prescriptive here in the library. To some extent, I'm a bit inclined to just let the caller use whatever value they want, even if it's what _I_ would consider excessively small or large, and not validate at all beyond ensuring it's a number (and defaulting the value if it isn't, rather than throwing).
> 
> If we're going to have a limit, I'd probably leave it as you have it.

I think most if not all would want to use 4, 8, 16 or 32 as alternatives to 64. So I think a limit makes sense.

src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
avrecko: I also first thought about putting it in NuProcess but it is an interface. Not a big fan of implementation in interfaces. Constants feels good to me. But if you like. No problem for me to put something in NuProcess.

avrecko: For more fine grained control I'd put it in the `NuProcessBuilder`.

`NuProcessBuilder(....).setStdoutBufferCapacity(4092).setStdinBufferCapacity(32000).setStdErrBufferCapacity(16000);`

src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java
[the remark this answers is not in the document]
bturner: Exceptions from static initializers on what _should_ be a recoverable "error" (like setting 512 when 1024 is the minimum--which is 100% arbitrary from the perspective that the code would actually work just fine with 512 byte buffers) is a poor consumer experience in my opinion. I think clamping makes more sense. And if the value is outright invalid (NaN), then the consumer just gets the default experience, which has worked fine for every consumer of the library up to this point (given it's been hard-coded until now), until they fix their setting.

I do think we need to get some logging in to report the misconfiguration. `java.util.logging` is what other classes use, so we should probably update `Constants` to have its own `Logger`, per normal practice, and log when the provided value is invalid.

PR #153, opened 2024-02-13
Made thread name less generic and more specific to NuProcess.

avrecko:
Seeing just "ProcessQueueN" show up in Linux system reports. It might be a bit confusing with all other system and application threads. With a more specific name it is easy to recognize this as a thread of NuProcess from looking at Linux logs and reports.


bturner:
Thanks for the improvement, @avrecko. I've updated your branch to also update the processor thread name on Windows.

PR #156, opened 2024-05-01
Alternative fix for linux race condition between onStart and onExit.

avrecko:
Continuing from #149.

Moved onStart processing to event processing thread. This way we make sure onStart is called first before any other EPoll event processing is done.

@bturner This is just my suggestion. Curious how you would fix it.

### DIFF
--- a/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
+++ b/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java
@@ -37,6 +37,8 @@ public class LinuxProcess extends BasePosixProcess
 {
    private final EpollEvent epollEvent;
 
+   private boolean startCalled;
+
    static {
       LibEpoll.sigignore(LibEpoll.SIGPIPE);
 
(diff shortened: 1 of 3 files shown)


bturner:
@avrecko,

Thanks for the idea, but I'd prefer to keep it where the process manages `onStart`. The approach I'm leaning toward splits the existing `IEventProcessor.registerProcess(T)` method to extract a new `queueRead` (sibling to the existing `queueWrite`). With the split, `registerProcess` does the `pidToProcessMap` and `fildesToProcessMap` updates and `queueRead` does the `epoll_ctl` calls for `stdout` and `stderr` (on Linux; it does the `kevent` calls on macOS). This allows the `LinuxProcess` code to call `registerProcess` where it currently does, and then do a `myProcessor.queueRead(this)` after `callStart`.

I've opened pull request #158 with that approach, and I've tested it on all of the platforms I can.

avrecko:
@bturner 

Makes sense to do it this way. Keeping the threading model as is.

Looks good. I'd probably name the new method `registerRead` or something similar.

I trust the tests are now run with the build process and catch this case.

PR #158, opened 2024-05-05
Prevent onStart/onExit race for fast-exiting processes.

bturner:
Immediately after a `LinuxProcess` is registered with `ProcessEpoll`, it's possible for soft exit detection to detect process termination. If a process exits quickly, this means it's possible for `NuProcessHandler.onExit` to be called before `onStart`. When that happens, `BasePosixProcess.callStart` throws a `NullPointerException` because exit cleanup clears the process handler.

The top of such a call stack looks something like this:
```
May 01, 2024 1:38:24 AM com.zaxxer.nuprocess.internal.BasePosixProcess callStart
WARNING: Exception thrown from handler
java.lang.NullPointerException
	at com.zaxxer.nuprocess.internal.BasePosixProcess.callStart(BasePosixProcess.java:587)
	at com.zaxxer.nuprocess.linux.LinuxProcess.start(LinuxProcess.java:80)
	at com.zaxxer.nuprocess.linux.LinProcessFactory.createProcess(LinProcessFactory.java:40)
	at com.zaxxer.nuprocess.NuProcessBuilder.start(NuProcessBuilder.java:260)
```

This change splits the existing `IEventProcessor.registerProcess` method into two: `registerProcess`, and `queueRead` (a parallel to `queueWrite`). `registerProcess` sets up mappings (`pidToProcessMap`, `fildesToProcessMap`), and then `queueRead` does the `epoll` (or `kevent`, on macOS) operations.

This split ensures the process is registered before `onStart` is called, which is necessary to allow process handlers to call `wantWrite` during `onStart`, while also ensuring `stdout`/`stderr` aren't consumed until after `onStart` is called (which, in turn, prevents early soft-exit detection).

- Added `IEventProcessor.queueRead(T)`
  - Implementing classes now map the PID in `registerProcess` (as appropriate for the host OS), and then set up `epoll`/`kevent` state in `queueRead`
  - For synchronous processes (`NuProcess.run`), `queueRead` is called by the `ProcessEpoll`/`ProcessKqueue` constructors. There is no risk of a race because the process classes don't call `IEventProcessor.run` (which actually consumes events) until after `onStart` has been called
- Updated version to `2.1.0-SNAPSHOT` because a new interface method has been added
- Unrelated: Updated Maven to 3.9.6 (if using `mvnvm`)

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -25,7 +25,7 @@
         <dependency>
             <groupId>com.zaxxer</groupId>
             <artifactId>nuprocess</artifactId>
-            <version>2.0.8-SNAPSHOT</version>
+            <version>2.1.0-SNAPSHOT</version>
         </dependency>
 
         <dependency>
--- a/mvnvm.properties
+++ b/mvnvm.properties
@@ -1,1 +1,1 @@
-mvn_version=3.8.5
+mvn_version=3.9.6
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>com.zaxxer</groupId>
     <artifactId>nuprocess</artifactId>
-    <version>2.0.8-SNAPSHOT</version>
+    <version>2.1.0-SNAPSHOT</version>
     <packaging>bundle</packaging>
 
     <name>NuProcess</name>
--- a/src/main/java/com/zaxxer/nuprocess/internal/IEventProcessor.java
+++ b/src/main/java/com/zaxxer/nuprocess/internal/IEventProcessor.java
@@ -54,6 +54,18 @@ public interface IEventProcessor<T extends BasePosixProcess> extends Runnable
    void registerProcess(T process);
 
    /**
+    * Queues read handling for the process's STDOUT and STDERR streams.
+    * <p>
+    * Prior to 2.1, this was performed by {@link #registerProcess}, but it was moved to a separate method
+    * to help avoid a race condition when starting a new process where soft exit detection could result in
+    * a fast-running process exiting before start was called on its process handler.
+    *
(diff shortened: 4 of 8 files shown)


bturner:
@wkritzinger-atlassian, if you have a moment, could you test this with Bitbucket?

I've verified it locally on Linux (x64 only) and macOS (x64 and aarch64, both on native hardware) and Windows (x64 only, not that there should be any risk of a regression on Windows here), so I know all the basics work, but I'd value the more thorough check.

wkritzinger-atlassian:
@bturner Apologies for the delay in my response. I ran all of our internal tests against the version that is in this change and it is all green.

I also had a look at this change and it seems good as well. I'll attempt to reproduce our zombie process problem that we're still having and keep you updated when I find something.

bturner:
No worries, @wkritzinger-atlassian. I appreciate you taking the time to run through it. I'm going to look at a couple other pull requests and then see if I can work through a 2.1 release. Even if it doesn't fix your zombie process issue (and if I'm honest I suspect it won't), at least it will clean up the `NullPointerException`s on racy starts for some of the faster `git` processes.

fanf:
Hello, 
Do you know when a release with that correction will be cut ? We believe we saw it two times IRL now, but it is extremely hard to test/reproduce. Having a version with the patch would help decide if it's the same problem.

fanf:
@bturner I was able to reproduce the `Exception thrown from handler: Cannot invoke "com.zaxxer.nuprocess.NuProcessHandler.onStart(com.zaxxer.nuprocess.NuProcess)" because "this.processHandler" is null` error (by pure luck) on my local system. 
I can confirm that the current master with that PR makes the message disappear.

fanf:
Thank for 3.0.0 release :heart_hands:

PR #160, opened 2024-12-14
Target Java 8 bytecode.

bturner:
This was overlooked as part of dropping Java 7 support in NuProcess 3.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,8 +174,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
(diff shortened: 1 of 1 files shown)


bturner:
Is this something you think should be done, Brett? Without this, aside from documentation, I don't think we've actually dropped support for Java 7 in practice.

If you do agree with this change, which release should we target it for? 3.0.1? 3.1.0? 4.0.0? Since you documented Java 7 support as removed in 3.0.0, I think it would be fair to consider this a bugfix and release it in 3.0.1.

bturner:
Also, is there any chance of getting the CircleCI build working again? Or disabling it if not? Having every commit and pull request report failed builds makes changes feel unsafe. (I'd be happy to dig into trying to get something working, but I lack the permissions necessary to do so.)

PR #163, opened 2025-06-05
Explicit JPMS support

vietj:
Motivation:

The library supports JPMS providing an automatic module name.

While this library aims to target Java 7, it remains possible using multi-release jar to support explicit JPMS modules.

Changes:

Update the build to use the module-info Maven jar which creates a module-info.class descriptor.

Update jna dependency to use jna-jpms which also is a multi release jar with a module-info descriptor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
     <dependencies>
         <dependency>
             <groupId>net.java.dev.jna</groupId>
-            <artifactId>jna</artifactId>
+            <artifactId>jna-jpms</artifactId>
             <version>5.11.0</version>
         </dependency>
 
@@ -167,6 +167,11 @@
                     <artifactId>nexus-staging-maven-plugin</artifactId>
                     <version>1.6.12</version>
                 </plugin>
+                <plugin>
+                    <groupId>io.github.dmlloyd.module-info</groupId>
+                    <artifactId>module-info</artifactId>
+                    <version>2.0</version>
+                </plugin>
             </plugins>
         </pluginManagement>
         <plugins>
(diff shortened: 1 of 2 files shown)


src/main/java/module-info.yml
```diff
@@ -0,0 +1,10 @@
+name: com.zaxxer.nuprocess
+requires:
+  - module: com.sun.jna
+exports:
+  - package: com.zaxxer.nuprocess
+  - package: com.zaxxer.nuprocess.codec
+  - package: com.zaxxer.nuprocess.internal
```
bturner, line 7: I think `internal`, `linux`, `osx` and `windows` should likely not be exported. They're all intended to be implementation details of using the API defined in `com.zaxxer.nuprocess`. For whatever use case(s) you're trying to support, is there a reason you're using those packages?

pom.xml
```diff
@@ -68,7 +68,7 @@
     <dependencies>
         <dependency>
             <groupId>net.java.dev.jna</groupId>
-            <artifactId>jna</artifactId>
+            <artifactId>jna-jpms</artifactId>
```
bturner, line 71: I'm mildly concerned about this change. Other codebases that leverage NuProcess may have dependency managed `jna`, but may or may not have an entry for `jna-jpms`. They'd either need to exclude it, or switch their configuration over, to avoid getting both `jna` and `jna-jpms` (and potentially at different versions).

I'm not sure there's any good way around this, here in NuProcess's POM, but at a minimum this seems like something we should at least mention in `CHANGES.md` to try and raise visibility for consumers.

